PR #119, opened 2023-07-13
chore: [INTEG-812] Add dependabot.yml file

whitelisab:
## Purpose

Adds a dependabot.yml file that mirrors the configuration we had in the apps repo.

## Approach

I kept the following configurations options under `ignore`: 
- Ignoring major semver updates
- Ignoring `@contentful/app-sdk` updates. It is configured this way in the apps repo, but I am open to changing it if we want.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+
+updates:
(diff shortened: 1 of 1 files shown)


ryunsong-contentful (approved):
Beautiful

PR #150, opened 2023-07-21
[INTEG-301] Allow 1000 max file limit per field for Cloudinary

ryunsong-contentful:
## Purpose
Support ticket request to increase the size https://contentful.atlassian.net/browse/INTEG-301

## Approach
Just update the constant

## Testing steps
Smoke test the app

## Breaking Changes
None

None

## Deployment
Standard


rusticpenguin (approved):
🫡

PR #157, opened 2023-07-24
[FIX] Format the index.js file and update the max file limit setup

ryunsong-contentful:
## Purpose
Small fix to the UI to make sure it says the correct max file limit number.

## Approach
Simple smoke test of the app

## Testing steps
Smoke test the app

## Breaking Changes
None

None

## Deployment
Standard


rusticpenguin (approved):
🎉

PR #162, opened 2023-07-26
feat: Cloudinary App rewrite

andipaetzold:
This is a rewrite of the Cloudinary app to allow more customizations in the future. It should behave identically to the app that's currently on the marketplace.


whitelisab (changes requested):
This rewrite looks really great! Thanks for making some improvements like converting to Typescript, adding tests, updating dependencies, adding Vite, etc. I confirmed that the app works as expected and that the underlying data is persisted in the same format.

I made some small comments, and a few in particular that should be addressed:
- There are a few linting errors (just need to remove some unused variables)
- I am suggesting that you remove the two deploy scripts. Then we can merge this in without having to worry about this deploying to our marketplace app for now. Does that work for you?

andipaetzold:
@whitelisab Thanks for the quick review!

I applied review comments. Removing the deploy scripts is fine - let's coordinate deployment internally.

whitelisab (approved):
Thanks for making the changes, looks great. Yes, let's coordinate deployment later, but for now, this is good to go!

PR #184, opened 2023-08-10
feat(cloudinary2): UI improvements

andipaetzold:
* use F36's AssetCard for asset thumbnails
* add additional metadata to the three-dot-menu

image


apps/cloudinary2/src/locations/Field/Thumbnail.tsx
```diff
@@ -76,7 +123,7 @@ function getUrlFromAsset(installationParams: AppInstallationParameters, asset: C
     api_key: installationParams.apiKey,
   });
 
-  const transformations = `${asset.raw_transformation ?? ''}/c_fill,h_100,w_150`;
+  const transformations = `${asset.raw_transformation ?? ''}/c_crop,h_149,w_194`;
```
whitelisab, line 126: I think that this could be tweaked a bit more, as it's hard to tell what some of the sample images are in the Thumbnail. For example, here is the same image in the original app vs. these new changes:

Screenshot 2023-08-15 at 9 02 58 AM

Screenshot 2023-08-15 at 9 02 51 AM

apps/cloudinary2/src/locations/Field/index.tsx
```diff
@@ -8,7 +8,7 @@ import { Thumbnails } from './Thumbnails';
 
 const Field = () => {
   const sdk = useSDK<FieldAppSDK<AppInstallationParameters>>();
-  useAutoResizer();
+  useAutoResizer({ absoluteElements: true });
```
whitelisab, line 11: I wasn't aware of this option when using `useAutoResizer`!

apps/cloudinary2/src/locations/Field/Thumbnail.tsx
```diff
@@ -61,12 +69,51 @@ export function Thumbnail({ asset, isDisabled, onDelete }: Props) {
...
+            <dl className={styles.fileInformation.dl}>
+              <dt>Location:</dt>
+              <dd>{asset.public_id.split('/').slice(0, -1).join('/') || 'Home'}</dd>
+              <dt>Format:</dt>
+              <dd>{asset.format}</dd>
+              <dt>Size:</dt>
+              <dd>{fileSize(asset.bytes).human('jedec')}</dd>
+              <dt>Dimensions:</dt>
+              <dd>
+                {asset.width} x {asset.height} px
```
whitelisab, line 106: Small nit: designs do not have a unit specified for dimensions, maybe remove the `px` ?

apps/cloudinary2/package-lock.json
whitelisab, line 1: There's some conflicts in `package-lock.json` and `package.json` that will need to be resolved

whitelisab (commented):
The UI improvements are awesome and the code looks great! 🚀 I had a few small comments, the most important is that the thumbnail image is a little too zoomed in and it's hard to tell what the selected image is.

apps/cloudinary2/src/locations/Field/Thumbnail.tsx
andipaetzold: Seems like the designs are updated 🤔  Josh asked me to add the unit.

apps/cloudinary2/src/locations/Field/Thumbnail.tsx
andipaetzold: Oh, yes that looks wrong.
Plan was to scale down and only crop one dimension. Basically css' `object-fit: cover`

PR #191, opened 2023-08-16
feat: Cloudinary backend

andipaetzold:
* Create new uuid every time the api secret changes
* Installation uuid is stored as installation parameter


apps/cloudinary2/src/locations/ConfigScreen/index.tsx
```diff
@@ -65,19 +65,21 @@ const ConfigScreen = () => {
   const [selectedFields, setSelectedFields] = useState<SelectedFields>({});
 
   const onConfigure = useCallback(async () => {
+    let installationUuid = parameters.installationUuid;
+    if (backendParameters.apiSecret.length > 0) {
+      installationUuid = window.crypto.randomUUID();
```
lilbitner, line 70: do we need to generate this new `UUID` if we default the value on line 98?

lilbitner (approved):
Code looks great 👍 just one small comment

apps/cloudinary2/src/locations/ConfigScreen/index.tsx
lilbitner: answered my own question looking at it 😅  => need to create a new UUID every time the apiSecret changes. I know that we didn't follow this with the Slack app fix, but is there a way to do a look-up via the existing `installationUuid` to then update the apiSecret value instead of creating a new entry?

andipaetzold: The "API secret" field is always empty when the config screen is loaded. We only create a new uuid when something was typed in and we have a new api secret.

If the user changes other fields, the "API Secret" field stays empty and we keep the old uuid.

lilbitner: ah gotcha, makes sense 👍

PR #197, opened 2023-08-17
[INTEG-1083] - Extend Bynder response with additional fields 

matthew-gordon:
## Purpose

In response to customer request, we are extending the `FIELD_SELECTION` query to add `textMetadataproperties`

## Approach

- adds new selection set to es6 (graphql) query


apps/bynder/src/index.js
```diff
@@ -5,6 +5,8 @@ import logo from './logo.svg';
 
 const CTA = 'Select a file on Bynder';
 
+// DOCS POINT TO THIS
+// const BYNDER_BASE_URL = 'https://ucv.bynder.com';
 const BYNDER_BASE_URL = 'https://d8ejoa1fys2rk.cloudfront.net';
```
matthew-gordon, line 10: The documentation points to the commented out `BYNDER_BASE_URL`, is there any reason known to the direct CDN reference?

.lintstagedrc.json
```diff
@@ -1,10 +1,4 @@
...
-    "eslint"
-  ],
-  "*.ts": [
-    "prettier --write",
-    "eslint"
-  ]
-}
\ No newline at end of file
+  "*.js": ["prettier --write"],
+  "*.ts": ["prettier --write"]
```
matthew-gordon, line 3: Lint stage was failing, needs to be addressed in another PR, removing this for now but happy to add it back and address this here if we think its necessary.

apps/bynder/src/index.js
```diff
@@ -57,6 +60,10 @@ const FIELD_SELECTION = `
   createdAt
   files
   originalUrl
+  textMetaproperties {
```
jsdalton, line 63: Only question I have on this PR is the fields below (name, value). Do we know they always appear in the graphql schema for all customers?

jsdalton (approved):
No complaints from me 👍

apps/bynder/src/index.js
matthew-gordon: Yeah the way this is declared in the GraphQL schema from Bynder:

```graphql
type TextMetaproperty {
  name:  String!
  value:  String!
}

type Query {
  ...
  textMetaproperties: [TextMetaproperty!]!
  ...
}
```

The bang operator `TextMetaproperty!` indicates that every returned type must have a `name, value` present, the bang operator trailing the array declaration `[TextMetaproperty!]!` indicates that every entity returned must return required fields, otherwise it will return an empty array by default.

apps/bynder/src/index.js
matthew-gordon: any thoughts here @jsdalton?

PR #234, opened 2023-09-01
[INTEG-1152] Fix: Bynder horizontal scroll

whitelisab:
## Purpose

When there are a lot of filters in the filter bar for the Bynder UCV, the whole dialog was scrolling horizontally and you couldn't see all of the features in the search bar above the filters. 

## Approach

Added `overflow-x:auto` to the container div.

Before:
Screenshot 2023-09-01 at 1 00 28 PM

After:
Screenshot 2023-09-01 at 1 00 17 PM

## Testing steps

Note that I could not add additional filters via our Bynder testing environment, so I added additional filters by creating new button elements in Dev Tools in order to replicate the issue. I also tested that adding the CSS property did not impact the Bynder UCV when there aren't too many filters in the top bar.

### DIFF
--- a/apps/bynder/src/index.js
+++ b/apps/bynder/src/index.js
@@ -82,7 +82,7 @@ function makeThumbnail(resource) {
 function prepareBynderHTML() {
   return `
     <div class="dialog-container">
-      <div id="bynder-compactview" />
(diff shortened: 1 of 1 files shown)


rusticpenguin (approved):
Nice!

PR #262, opened 2023-09-12
feat(Cloudinary): switch to Cloudinary v2

anho:
## Purpose

Preparing release.


andipaetzold (approved):

clarkgunn:
For those that use Cloudinary as the source of truth for alt text, would be great to view and edit that here.

PR #272, opened 2023-09-14
feat: create Surfer app

marcinwysocki:
## Purpose

Integration with Surfer's guidelines extension, to enable SEO optimisation of content created in Contetful.

## Approach

This app embeds Surfer guidelines (previously available i.e. as a Chrome extension for Google Docs) in Contentful using three locations:

- App configuration as an entrypoint and a place to configure content types and fields that Surfer should be enabled on
- Sidebar for the main functionality
- Dialog for configuration, as it's much wider than the guidelines view

Because the guidelines are based on HTML content, they can be calculated only based on RichText fields. In case there's more of those in the content model, there's an option to choose which ones should be included.

The extension itself uses our pre-existing iframe-based solution.   

## Testing steps

Testing the workflow below requires a Surfer subscription. Please message me at marcin.wysocki@surferseo.com after you create a test account and I can sort it out for you.

1. Install the Surfer app and add it to a content model containing RichText field(s)
2. Sign in to Surfer.
3. Either create a new Content Editor or choose an existing one 
4. If there are multiple RichText fields in the content model, choose which ones should be taken into consideration for Surfer guidelines
5. Start writing. The Content Score and other guidelines should update accordingly.

## Breaking Changes

None

We've discussed this integration with Matt Middleton.

## Deployment

No special requirements.


gavinmatt (approved):
Looks great-  thanks for the updates!

jsdalton (approved):

jsdalton:
Thanks for the PR @marcinwysocki. In preparation for launch I'm going to merge this PR now, so I we can add a few extra elements needed for deploy and release.

PR #277, opened 2023-09-15
fix(Cloudinary): store API Secret after installation

anho:
## Purpose

When installing for the first time we need to wait for the installation to finish in order to utilize signed requests.


andipaetzold:
I explicitely put the saving of backend parameters into the `onConfigure`. So, in case saving the backend params fails, we don't store an orphaned installation uuid.

anho:
@andipaetzold is there another way to deal with the first installation scenario?

andipaetzold:
@anho I can't come up with one... 🤔  Let's do it like this then 😞

andipaetzold (approved):

PR #278, opened 2023-09-15
fix(Cloudinary): prevent flickering when opening dialog

anho:
## Purpose

When opening the menu from the Thumbnail the iframe is flickering. Working some CSS magic to fix that.


andipaetzold (commented):
Shouldn't we move this to the dialog component then?

Also, what kind of flickering do you see?

andipaetzold (approved):

PR #290, opened 2023-09-19
fix: adjust Bynder dialog width [INTEG-1305]

whitelisab:
## Purpose

Reported bug in the Bynder app where when you use the search and only a few assets are displaying (so there are fewer columns), the content in the dialog shrinks in width.

## Approach

Adjusted CSS for the `bynder-compactview` div to make `width: 100%`. See before and after below:

Before:
Screenshot 2023-09-19 at 7 55 11 AM

After:
Screenshot 2023-09-19 at 7 55 19 AM

## Testing steps

1. Go to a space where the Bynder app is installed
2. Go to an entry and navigate to a field that's using the Bynder app
3. Click on "Select a file on Bynder" button
4. In the dialog, enter something in the search field that limits results to only a few assets
5. Confirm that the width of the content in the dialog doesn't change when there are only a few items displayed

## Breaking Changes

None

### DIFF
--- a/apps/bynder/src/index.js
+++ b/apps/bynder/src/index.js
@@ -82,7 +82,7 @@ function makeThumbnail(resource) {
 function prepareBynderHTML() {
   return `
     <div class="dialog-container">
-      <div id="bynder-compactview" style="overflow-x:auto;" />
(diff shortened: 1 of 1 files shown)


rusticpenguin (approved):
1 line, large impact 🥳

PR #296, opened 2023-09-21
Add FlexFields by Thrillworks

vipulchadda:
## Purpose

FlexFields™ by Thrillworks adds an additional layer of flexibility to the content models by allowing fields to be displayed based on the the values in the content entry. This allows the reuse of content types for minor variations, as well as content entries. FlexFields not only hides fields in the same entry, but also on referenced entries (child entries) providing more flexibility.
One of the main reasons for build this app was to solve for better author experience when using layout pattern for content modelling where we have the data entry (Topic) that holds the data and the component (Assembly) that defines the layout of the component. This content model works really well with UI and FE development, but without FlexFields, is very difficult to manage from content authoring perspective. With this app, we aimed to improve this authoring experience.

## Approach

FlexFields adds a new entry editor for the configured content types to hide fields based on the rules in the configuration. The entry editor uses the default field editors to show the fields that are not hidden.
The app has the following locations:
**ConfigScreen** - Screen to configure rules to hide fields. The entry editor is updated automatically as the default entry editor for all the content types added in the configuration (both the driving content types as well as the affected content types).
flexible-fields-add-rule

**EntryEditor** - Add a new entry editor tab which displays the fields conditionally for the respective entry, based on the rules set by the user
FlexFields-EntryEditor

## Testing steps

1. Apps → FlexFields → Configure
2. Fill out all the fields to create a rule and Save → Verify the new rule is visible in the ‘Current Rules’ list below
3. Content tab → Open any content entry for which the rule was created
4. Update the content entry to satisfy the rule that was created in the config screen
5. Make sure, the target field in the rule is hidden


gavinmatt:
Did a functional test - looks great!

gavinmatt (approved):

jsdalton (approved):

jsdalton:
Thanks for the PR @vipulchadda . In preparation for launch I'm going to merge this PR now, so I we can add a few extra elements needed for deploy and release.

PR #335, opened 2023-10-02
Fix/cloudinary localization bug

jsdalton:
## Purpose

We received customer reports that inserting images was broken for localized entries. When inserting an image for one locale it was also overriding the "default" locale.

Bug seems to related to an erroneous logic in the `useFieldValue` hook here https://github.com/contentful/ui-extensions-sdk/blob/fa8a06f9d6ded33fcef2b8cad87fce7b2cb2c3e2/lib/field.ts#L87 which assumes that the _default_ locale (`sdk.locales.default`) should be used when a locale is not specified, which contravenes the behavior of `field.setValue()` which uses the _current_ field locale (ie. `sdk.field.locale`) if one is not explicitly provided.

## Approach

* Just pass the correct locale value explicitly

## Testing steps

Reproduced bug locally then verified the fix manually (no tests in this part of the app):

image

### DIFF
--- a/apps/cloudinary2/package-lock.json
+++ b/apps/cloudinary2/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "cloudinary2",
+  "name": "@contentful/cloudinary-assets",
   "version": "2.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


mgoudy91 (approved):
🔥

PR #345, opened 2023-10-03
feat: new default selected asset displays asset name [INTEG-1405]

whitelisab:
## Purpose

[This PR](https://github.com/contentful/apps/pull/5020) updated the `dam-app-base` with a new default for when there is no url/thumbnail image for an asset. This current PR bumps the `dam-app-base` in the Bynder app to utilize this change. It also changes the value of the alt text for the thumbnail to display the asset name instead of the asset id. Many of our other DAM apps return the asset name or a more readable Id from the `makeThumbnail` function.

## Approach

The asset name is a friendlier thing for users to see when there is no thumbnail image to display. 

Before:
Screenshot 2023-10-03 at 1 11 35 PM

After:
Screenshot 2023-10-03 at 1 12 15 PM

## Testing steps

Select an asset from Bynder that doesn't have a thumbnail image to see the new default asset in the entry. 

## Breaking Changes

None, this is purely a visual change.

### DIFF
--- a/apps/bynder/package-lock.json
+++ b/apps/bynder/package-lock.json
@@ -8,7 +8,7 @@
       "name": "@contentful/bynder-assets",
       "version": "1.11.16",
       "dependencies": {
-        "@contentful/dam-app-base": "^2.0.35",
+        "@contentful/dam-app-base": "^2.1.0",
         "react": "16.14.0",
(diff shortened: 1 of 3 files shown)


matthew-gordon (approved):
👍 🚀

PR #354, opened 2023-10-05
ci: remove unneeded vault access

jsdalton:
## Purpose

Partners are seeing build fails because of authorization issues with CircleCI checks. EG https://app.circleci.com/pipelines/github/contentful/marketplace-partner-apps/706/workflows/cbd48495-086f-4e4f-a384-87f3d31bcae2

image

## Approach

Hard to know exactly what's causing this but I suspect it's the vault requirement that's hitting an auth problem. We actually don't even need access to the CI vault for any of our tests, so I want to try removing it. (If we do end up needing this access in CI I'll find out the hard way).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,24 +2,6 @@ version: 2.1
 
 orbs:
   node: circleci/node@5.0.2
-  vault: contentful/vault@1
-
-commands:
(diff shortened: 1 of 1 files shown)


lilbitner (approved):
👍

PR #366, opened 2023-10-09
fix: menu item display for non-image assets [INTEG-1415]

whitelisab:
## Purpose

With the refreshed Cloudinary app, users can see additional file information in a menu for selected assets. Non-image assets don’t have format or dimensions data, so currently the menu information doesn’t look right for these assets.

## Approach

Only show the format and dimensions data when it exists. See screenshots below.

Before:
Screenshot 2023-10-09 at 9 40 37 AM

After:
Screenshot 2023-10-09 at 9 40 12 AM

## Testing steps

Select a non-image asset in Cloudinary. The file information in the menu should only show location, size, and created date. Any image or video asset should also display the format and dimensions.


lilbitner (approved):
Nice 🙌

rusticpenguin (approved):
Looks good!

PR #475, opened 2023-11-20
chore: release image hotspot creator

rusticpenguin:
## Purpose

Release Image Hotspot Creator to the production app definition

## Approach

Had to fix a merge conflict with package locks being out of sync. From discussions with the team we have decided to go with main as the source of truth here


README.md
```diff
@@ -38,7 +38,6 @@ In order for your app to be deployed successfully (with the exception of the tes
 When you are ready to submit your app to the [marketplace apps repo](https://github.com/contentful/marketplace-partner-apps), clone the repo locally, create a new branch with a descriptive name for the app/changes you are introducing, and create a pull-request against the `main` branch. An ecosystems integrations will review your PR soon and let you know if any changes or clarifications are needed. To ensure the fastest approval time on your PR, make sure you:
 
 - PR titles must confirm to the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) standard. (A Github Action will emit a failing status check if your PR title does not match)
-- Add the new app to the release-please-config.json file so that the release workflow will include the new app. The release workflow encapsulates versioning as well as the creation and maintenance of a changelog,
```
lilbitner, line 41: I think we probably want to keep this added text here!

lilbitner (commented):
hmm the more that I look at this the more I am not sure this has to do with versioning (the app version is the only thing we changed) - it looks like we are downgrading a lot of packages. What are your thoughts?

README.md
rusticpenguin: Oops I ran my script for all of these backwards! Was supposed to accept incoming. I'll fix this tomorrow morning!

lilbitner (approved):

PR #557, opened 2023-12-06
feat: display additional asset info and preselect assets in modal [INTEG-1611]

whitelisab:
## Purpose

This change makes two updates to the Bynder app to improve the user experience and make it easier to know which assets are selected:

1. Displays a "More details" section in the asset card that displays asset name and id
2. Creates a configuration option to prefill selected assets in the modal (which also changes the save behavior after selecting assets in the modal)

## Approach

This is a follow up after the changes in the `dam-app-base`, so [this PR](https://github.com/contentful/apps/pull/5682) passes in functions for the two new options.

More details display:
Screenshot 2023-12-06 at 12 10 15 PM

Prefill config option:
Screenshot 2023-12-06 at 11 44 23 AM

## Testing steps

Bynder development app should display the more details section and a new config option on the config screen. When selecting that option, the selected assets should be prefilled in the picker.

## Breaking Changes

No breaking changes, the prefill option default is existing behavior. The new "more details" section will show for all assets.

### DIFF
--- a/apps/bynder/package-lock.json
+++ b/apps/bynder/package-lock.json
@@ -8,7 +8,7 @@
       "name": "@contentful/bynder-assets",
       "version": "1.0.1",
       "dependencies": {
-        "@contentful/dam-app-base": "^2.1.0",
+        "@contentful/dam-app-base": "^2.2.0",
         "react": "16.14.0",
         "react-dom": "16.14.0"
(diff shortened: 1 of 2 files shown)


rusticpenguin (approved):
Looks good! 🚢

PR #600, opened 2023-12-14
feat: First commit, adding Lilt app to Contentful Marketplace

the-julianson:
## Purpose

First commit to move our Lilt extension to the Contentful Marketplace.

## Approach

Follow the steps in the main README.md of the current repository.


whitelisab (changes requested):
Hi! The app looks great, I confirmed that the various scripts work as intended.

Given that your app is an existing app in the marketplace and we need to transfer ownership of the app definition, could you please remove the two deploy scripts for now? I will then merge in this PR, and then after we transfer the production app definition to Contentful, I will open a PR to add those scripts back in. I will also configure this app to update a `CHANGELOG` in a follow up PR. Let me know if you have any questions!

whitelisab (approved):
I removed the deploy scripts for now

gavinmatt (approved):

apps/lilt/src/emc-dialog.js
```diff
@@ -0,0 +1,337 @@
+/* global ManagementClient */
+import React, { useState, useEffect, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import { useAsync } from 'react-use';
+import { Stack, IconButton, Heading } from '@contentful/f36-components';
+import { CloseTrimmedIcon } from '@contentful/f36-icons';
+import CMClient from './contentful-management-client';
+
+const BASE_URL = process.env.CONNECTORS_BASE_URL;
```
whitelisab, line 9: These environment variables (`CONNECTORS_BASE_URL`, `HYDRA_CLIENT_ID`, and `HYDRA_CLIENT_STATE`) were likely configured in your repo where this app was hosted previously, but are not configured here.

whitelisab (changes requested):
@the-julianson I just noticed some environment variables in the code that are not configured, so we will need to find a way to address this to ensure that the app functions properly.

apps/lilt/src/emc-dialog.js
the-julianson: @whitelisab Yes, let me get back to you with those values. We were using Kubernetes to handle the env vars.

the-julianson:
@whitelisab 
First, thanks for your review!

I have a couple of questions:
Something I wanted to discuss with my team is about the commit history. For this PR I just copied the whole master branch from the repo in LILT to this one, and created a PR. This has deleted all the commit history, I saw other projects also having their history initialized in the first commit. Is it possible to use REBASE or EXPORT/IMPORT strategies to maintain the commit history. 

Another question, how to handle the different environments. We were using branch tagging to trigger automatic deployment to our DEV (LILT Preview) and our PROD (LILT) instances.

PR #603, opened 2023-12-15
fix: resolve dependency graph conflicts

jsdalton:
## Purpose

Two of our apps (ceros and image hotspot creator) did not actually install locally. Somehow lerna is executing the installation with an implicit --force or --legacy-peer-deps 🤔 

## Approach

* Update dependencies to resolve issue


rusticpenguin (approved):
Looks good!!

Not needed here, but it would be nice to upgrade cero's index to use the new root api since we're upgrading them to v18

https://legacy.reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

PR #606, opened 2023-12-18
ci: move deploy to githubactions

jsdalton:
## Purpose

The current deploy strategy requires us to maintain a special "production" branch which gets merged into periodically. For reasons it's prone to merge conflicts. It's also separate from our release please process, which maintains and publishes the actual version releases and change logs.

It would be better if the production deploy could be coupled more closely to the release process vs. being a separate process we have to maintain and execute indepedently.

## Approach

* Remove circle ci deploy steps and reimplement in GH action
* Rename current release please workflow to better reflect what it's doing
* Bring in secrets required by deploy steps
* Run staging deploy on all builds, but production only when release is created ... that is when release please PR is merged
* Use `releases_created` output to trigger production deploy

## Testing steps

* See this test build which was executed on the PR and had deploy disabled. https://github.com/contentful/marketplace-partner-apps/actions/runs/7250038177?pr=606

See https://github.com/google-github-actions/release-please-action?tab=readme-ov-file#outputs


whitelisab (approved):
Looks good! 🌟

rusticpenguin (approved):
🚀

PR #614, opened 2023-12-19
ci: don't deploy to staging on release

jsdalton:
## Purpose

Since we deploy to staging on all merges to main, we don't need to redeploy the same thing on the release please PR merge.

## Approach

* Suppress staging deploy when releases are created

See https://docs.github.com/en/actions/learn-github-actions/expressions#operators


whitelisab (approved):
🏅

PR #625, opened 2023-12-20
chore: change DX codeownership to /apps/*/src & apps/*/public only

t-col:
change DX codeownership to /apps/*/src & apps/*/public only to reduce PR noise from dependency upgrades


whitelisab (approved):
LGTM! 🌟

PR #654, opened 2024-01-03
feat: upgrade shopify storefront API version

lilbitner:
## Purpose

Currently the Shopify Storefront API version that we are using, 01-2023, within the Shopify app has been deprecated as of January 1st, 2024. 
Screenshot 2024-01-03 at 12 12 30 PM.

User's of the Shopify app are now receiving messages that they are interacting with a deprecated API. We therefore need to update this. 

## Approach

I updated the API version to the latest version, in conjunction with a decision reached with the External References team, who will simultaneously update to the same version. The breaking changes that span the updates in the last year do not apply to our functionality. 

Question: How do we best keep track of these updates in the future? Since we don't receive these emails from Shopify, users do. A reminder? ticket?

## Testing steps

Test that the Shopify app still works locally. 

## Breaking Changes

To my knowledge after reading the changelog, there should be no breaking changes within the context of our use of this API.

### DIFF
--- a/apps/shopify/src/constants.js
+++ b/apps/shopify/src/constants.js
@@ -3,7 +3,7 @@
 // whether the returned variant is the default one or not
 export const DEFAULT_SHOPIFY_VARIANT_TITLE = 'Default Title';
 // upgrade API version to next stable version every quarter
-export const SHOPIFY_API_VERSION = '2023-04';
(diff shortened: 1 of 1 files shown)


mgoudy91 (approved):
🎉

BasKiers (approved):

Issue #642, opened 2023-12-28
[FlexFields] Field settings not rendering


perenstrom:
Relates to **FlexFields v1.1.1** (ping @rusticpenguin who seems to have committed to this app recently.)

**Describe the bug**
Field settings such as help text, boolean alternatives lables, etc. are not rendered when using the FlexFields entry editor.

**To Reproduce**
1. Create a content type with a boolean field, where true-label and false-label has been changed. Also add a help text.
2. Add the FlexField entry editor to that content type
3. Create a content item with that type.

**Expected behavior**
All custom labels and help texts should be visible.

**Screenshots**
Boolean custom labels only show "yes" and "no", no help text is shown.

whitelisab:
Hi @perenstrom! Thank you for raising this issue and for the clearly documented steps to reproduce. I was able to see that the FlexFields entry editor does not carry over custom labels or help text from the original field. After looking at the code, I am not sure that this was something that worked in the app previously, had you experienced this working previously with the app? I am tagging our partner developer @vipulchadda from Thrillworks who manages this app. They can help verify if this is expected behavior for the app or whether this might be something that they will address in the future. Again, thank you for providing this feedback!

perenstrom:
@whitelisab Thanks for your response! This was the first time trying it out, so I don't know if it has worked previously.

vipulchadda:
Hi @perenstrom, thanks for raising this issue and thank you @whitelisab for tagging me and getting this to our attention. We are currently looking into this issue and will get back as soon as we have any update on this.

lilbitner:
Thank you for the update @vipulchadda and for the quick fix, which has now been merged to production. This issue should now be resolved in the app @perenstrom if you would like to confirm that it is working for you!

perenstrom:
@lilbitner @vipulchadda Thanks for the fix! That indeed fixes help texts and custom booleans, however there are still some problems. Let me know if I should create separate issues for these:

* Rich text field editor doesn't show any content at all (preview tab displays text ok)
* No support for multiple locales, only viewing default language regardless of selected locale
* Custom UI Extensions doesn't render at all
* Field headings are still light grey, instead of black as in normal editor (makes readability worse)
* Relation fields are in a different style (minor)

vipulchadda:
@perenstrom thank you for raising these issues. We will look into these and get back to you. I will let @lilbitner / @whitelisab decide if we need separate issues for these.
[...]

PR #685, opened 2024-01-09
fix: Field customizations and additional config scenarios

vipulchadda:
## Purpose

1. Fix for https://github.com/contentful/marketplace-partner-apps/issues/642 - Show field customizations on FlexFields entry editor
2. Fix rules configuration for child entities from single reference fields.
3. Fix rules configuration and hiding fields for references to same content type.

## Testing steps

1. Add help text for any field and/or custom true/false labels for boolean field in content type configured in FlexFields. It should now show up in the FlexFields entry editor.
Screenshot 2024-01-09 at 5 16 44 PM
2. Add reference field with single reference to content type configured in FlexFields. The content type(s) should be listed in the `Hide Field` section
3. Add reference field referencing the same content type. The entry editor should correctly hide fields in the same entry vs child entry based on the rule configured.

### DIFF
--- a/apps/flexfields/src/components/RulesList.tsx
+++ b/apps/flexfields/src/components/RulesList.tsx
@@ -166,6 +166,7 @@ const RulesList = (props: any) => {
                     {/* {rule.targetEntity} */}
                     {getContentTypeName(rule.targetEntity, allContentTypes) ??
                       rule.contentType}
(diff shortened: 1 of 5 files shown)


lilbitner (approved):
This looks great! Tested it locally, and all works well. Thank you for making these adjustments!

PR #832, opened 2024-02-08
chore: decouple dependabot updates per app

jsdalton:
## Purpose

When we introduced #810 we observed an unanticipated side effect: that dependabot was combining updates for the same dependency across multiple apps into a single PR. An example is here: https://github.com/contentful/marketplace-partner-apps/pull/814

Per our new guidance we will be asking app developers to approve major version updates to their dependencies individually. However, the grouping of PRs adds signficant friction to this process because we would require multiple reviewers on the same PR, and broken updates for a single app would hold up a change for all other updates in that PR.

## Approach

* Force dependabot to treat packages individually by splitting the udpate directories explictily
* Use yaml anchor and alias to avoid repetition
* There's an added benefit here where e.g. we can specify specific ignores per directroy

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,9 +2,9 @@ version: 2
 
 # note: we split each app out individually to force dependabot to open individual, separate PRs for each apps
 # so that app changes are decoupled from one another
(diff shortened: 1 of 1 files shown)


whitelisab (approved):
⭐

PR #866, opened 2024-02-08
chore: pin node types to match supported node version (v18)

jsdalton:
## Purpose

We have a lot of dependabot PRs trying to push up types to v20, eg https://github.com/contentful/marketplace-partner-apps/pull/864

Types should match the supported version of node which is `v18.17`

## Approach

* Pin to highest types package available for node 18


matthew-gordon (approved):
🎸

PR #884, opened 2024-02-09
chore: remove the outdated cloudinary from this repo

jsdalton:
## Purpose

A while back we replaced the "original" Cloudinary with a new improved version that supported external references (i.e. "Cloudinary 2". At the time we kept the original Cloudinary around, in case we needed to fallback to it: https://github.com/contentful/marketplace-partner-apps/pull/262

That time has long since passed and we can now safely remove the old Cloudinary and keep Cloudinary 2.

## Approach

* Remove the old cloudinary
* Follow up PR will rename (just separating to make things a bit less confusing)


whitelisab (approved):

andipaetzold (changes requested):
Should be removed here as well: 

* https://github.com/contentful/marketplace-partner-apps/blob/main/release-please-config.json#L5
* https://github.com/contentful/marketplace-partner-apps/blob/3b44a1b9052e0cc7548d02e9a90defad44d52520/.github/dependabot.yml#L67-L81

jsdalton:
@andipaetzold TY good catch! 😅

andipaetzold (approved):

Issue #702, opened 2024-01-12
[FlexFields] Rule set does not seem to work correctly


houtaiwong-toast:
Hello, I'm testing out your app in an test environment and I can't seem to get it to function correctly at all.

**Describe the bug**
Adding rules does not show and hide the correct fields when changing "layout".

**To Reproduce**
I've attached a video file of our screen recording of the issue. 

**Expected behavior**
When rules are added the correct fields show/hide when switching between "layout" field values. 

https://github.com/contentful/marketplace-partner-apps/assets/55763151/2e6b1c33-e5be-4e36-a01f-1341ba6e9a7a

lilbitner:
Hi @houtaiwong-toast, thank you for raising this issue! Currently a constraint with the app framework is that you must refresh the Contentful web app once a change to the app configuration has been made. Do you mind trying to refresh the web app after setting the rules on the configuration page, and letting me know if this resolves the issue? If not, we will certainly look into it.

vipulchadda:
@houtaiwong-toast It looks like not all fields are showing up in the FlexFields editor. Can you share how the content type is set up and also how this looks in the default Contentful editor so that we can reproduce and debug this further. Thanks.

sizzle168:
Not sure if it's the same issue but it at least looks similar to this one here.
I created 5 simple rules that check if a short text field contains a specific value and if so, hide another field. But these rules only work partly. When I look into the browsers dev tools I can see that https://api.contentful.com blocks some calls with HTTP 429 Too many requests. Could this be the reason and if so, what can I do about it?

sizzle168:
@lilbitner @vipulchadda Do you think that @houtaiwong-toast's problem is also because of the contentful rate limiting?

vipulchadda:
@sizzle168 I don't think there should be any issues with rate limiting. Would you be able to share more details about the fields you are using and the rules that are set. This will allow us to reproduce this issue and dig into this further.

sizzle168:
This is how the content type looks:
<img width="806" alt="Content Type" src="https://github.com/contentful/marketplace-partner-apps/assets/19364321/e41dc314-ccef-4948-9dab-b3b2c025f522">

These are the configured rules:
<img width="1310" alt="Rules" src="https://github.com/contentful/marketplace-partner-apps/assets/19364321/46fd7867-7291-40e1-8ecd-1c7df3053fbb">

[...]

PR #919, opened 2024-02-13
fix: Fix edit rule for same entry

vipulchadda:
## Purpose

Fix for https://github.com/contentful/marketplace-partner-apps/issues/702 

## Testing steps

Create a rule for hiding fields in the same entry. Then edit that rule and confirm that the rule remains for same entry

https://github.com/contentful/marketplace-partner-apps/assets/31496466/80457d7e-edb8-4171-ae64-330b98908ee3


whitelisab (approved):
Thanks for this fix, looks good to me!

Issue #931, opened 2024-02-15
[FlexFields] Fields disabled for editing are being shown


JinHao-L:
**Describe the bug**
Fields disabled for editing are still shown in FlexFields editor page

**To Reproduce**
1. Create a content model with a hidden field (`Disabled Editing field`)
    <img src="https://github.com/contentful/marketplace-partner-apps/assets/52820430/1655f4b7-ffe4-491e-91dd-c3db73163bb4" width="600">

3. `Disabled Editing field` is being shown on FlexFields editor page
    <img src="https://github.com/contentful/marketplace-partner-apps/assets/52820430/423fe66b-cbc6-458c-b912-34f98d232373" width="600">

4. Disabled field is not shown on default editor page
    <img src="https://github.com/contentful/marketplace-partner-apps/assets/52820430/2c048bcc-5617-45c6-b105-c788b81792ae" width="600">

**Expected behavior**
- Disabled field should be hidden on FlexFields editor page by default
- The hidden field should reappear when writer toggle hidden fields
    <img src="https://github.com/contentful/marketplace-partner-apps/assets/52820430/d17ce306-8276-45e0-a04a-fae8142369ec" width="300">

**Additional context**
I did an exploration of contentful SDK API and found that we are able to get the disabled state from SDK Field API [Reference](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#getting-disabled-state-of-field)
- `sdk.field.getIsDisabled()`
- `sdk.field.onIsDisabledChanged(callback)`

Not sure if this would be helpful in designing a solution

vipulchadda:
Thank you raising this and giving the additional context. We will look into this and get back to you once fixed.

vipulchadda:
@lilbitner would you be able to help with the PR raised to fix this issue (https://github.com/contentful/marketplace-partner-apps/pull/949)

whitelisab:
@JinHao-L I merged in @vipulchadda's fix and it is now deployed to production. Can you please verify that this issue is now resolved? Thank you!

JinHao-L:
@whitelisab @vipulchadda 
Hi, I have verified that the fix works great! Thanks alot for the prompt fix 👍 

Will close this issue

PR #949, opened 2024-02-21
fix: Hide fields hidden for editing

vipulchadda:
## Purpose

Fix for https://github.com/contentful/marketplace-partner-apps/issues/931 

## Testing steps

Setup
- Set a field to be `Hidden when editing` for a content type. 
- Configure that content type to have a rule in FlexFields.

Result
- The field should be hidden in the FlexFields entry editor. 
- It should also show when selecting `Show hidden fields` in editor options.
- When creating a rule, the hidden field is called out in the options to hide field

Screenshot 2024-02-20 at 5 10 18 PM

https://github.com/contentful/marketplace-partner-apps/assets/31496466/14faf645-bfc5-4195-a5fd-d583f29336a1

Screenshot 2024-02-20 at 5 10 10 PM

### DIFF
--- a/apps/flexfields/src/locations/ConfigScreen.tsx
+++ b/apps/flexfields/src/locations/ConfigScreen.tsx
@@ -665,16 +665,16 @@ const ConfigScreen = () => {
                     )}
                     className={css({ width: 300 })}
                   >
-                    {targetEntityFields.map(({ id, name }: any) => {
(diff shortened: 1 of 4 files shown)


whitelisab (approved):
This fix looks good, I tested that it works as expected. Thank you @vipulchadda!

PR #1092, opened 2024-03-14
fix: resolve emotion update in Cloudinary

lilbitner:
## Purpose & Approach

There is a major version update [PR](https://github.com/contentful/marketplace-partner-apps/pull/913) to Cloudinary that cannot be resolved given the lack of TypeScript support for `@emotion/core`. The CI error points to the lack of a TS declaration file for @emotion/core. After digging around to see how we can support `@emotion/core`, it seems that `@emotion/core` ultimately is really a suboptimal choice to install for use of emotion in a React app. 

The emotion [docs](https://emotion.sh/docs/introduction) do not explicitly support or encourage this package, and @emotion/core as a [package](https://www.npmjs.com/package/@emotion/core) was last published 2 years ago. There are also countless threads encouraging others to transition to @emotion/react or another supported package depending on use-case. 

The @emotion/react [package](https://www.npmjs.com/package/@emotion/react) supports the same functionality, with full support for TypeScript. Therefore I have replaced @emotion/core with @emotion/react in the Cloudinary package.

## Testing steps

I tested that Cloudinary locally still functions and works the same with this change.


mgoudy91 (approved):
👍

matthew-gordon (approved):

PR #1133, opened 2024-03-16
fix: Uploadcare app

igoradamenko:
## Purpose

Currently Uploadcare app is broken due to dependabot's autobump of @uploadcare/blocks dep. 

You see, [according to semver](https://semver.org/#spec-item-4) when a package is v0 it means that any update potentially has breaking changes:

> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

This is exactly [the state of @uploadcare/blocks](https://github.com/uploadcare/blocks/blob/main/CHANGELOG.md). So, during one of those autobumps by dependabot, Blocks lib was updated and the app became broken.

Uploadcare users reported the issue, so here we are, fixing it in this PR along with some deps updates.

## Approach

We've updated Blocks to the latest version and migrated the code accordingly. 

We could roll the Blocks version back to the working one without any code changes, but it would be harder to update in the future then :-)

## Testing steps

The issue users reported was related to fact that the file uploading window reset its own state when use clicked “Done”.

So, to test that everything works fine now you need to:

1. Setup the app.
2. Try to upload any image.
3. If the image is uploaded, then everything works fine.

## Breaking Changes

None.

[Blocks changelog](https://github.com/uploadcare/blocks/blob/main/CHANGELOG.md).

Also this PR probably closes #889 and #850.

## Deployment

—


lilbitner (approved):
Thank you for making these changes @igoradamenko. All looks great, I see there are a few merge conflicts on this branch, but once those have been adjusted, I am happy to merge this. And I will also look into adjusting the dependabot configuration for the uploadcare/blocks package, apologies for the issues that it caused.

lilbitner:
> @lilbitner, is there any estimate on release date? When should we test the app in production? 🙂

Hi @igoradamenko I have merged the changes to production just now, so you should see the changes in production within the next 10 minutes. I have also ensured that dependabot ignores updates to the @uploadcare/blocks dep in the future. Thank you for your patience and efforts here.

lilbitner:
@igoradamenko Wonderful news!!! 😃 🙌

PR #1550, opened 2024-04-17
fix: handle null fields in cloudinary

rusticpenguin:
## Purpose

Cloudinary was broken with the new SDK changes, this handles the SDK returning null

### DIFF
--- a/apps/cloudinary2/src/locations/Field/index.tsx
+++ b/apps/cloudinary2/src/locations/Field/index.tsx
@@ -18,26 +18,23 @@ const Field = () => {
   const sdk = useSDK<FieldAppSDK<AppInstallationParameters>>();
   useAutoResizer({ absoluteElements: true });
 
-  const [assets = [], setAssets] = useFieldValue<CloudinaryAsset[]>(sdk.field.id, sdk.field.locale);
-  const assetsIsNotNull = assets !== null;
+  const [fieldValue = [], setAssets] = useFieldValue<CloudinaryAsset[] | null>(sdk.field.id, sdk.field.locale);
+  const assets = fieldValue || [];
 
   const [editingEnabled, setEditingEnabled] = useState(!sdk.field.getIsDisabled());
   useEffect(() => {
     sdk.field.onIsDisabledChanged((disabled) => setEditingEnabled(!disabled));
   }, [sdk.field]);
 
-  const canAddAssets = assetsIsNotNull && assets.length < sdk.parameters.installation.maxFiles && editingEnabled;
+  const canAddAssets = assets.length < sdk.parameters.installation.maxFiles && editingEnabled;
 
   return (
     <>
       <Global styles={globalStyles} />
       <GlobalStyles />
       <Stack spacing="spacingM" flexDirection="column" alignItems="flex-start">
-        {assetsIsNotNull && assets.length > 0 && <Thumbnails assets={assets} onChange={setAssets} isDisabled={!editingEnabled} />}
-        <OpenDialogButton
-          isDisabled={!canAddAssets}
-          onNewAssetsAdded={(newAssets) => setAssets(assetsIsNotNull ? [...assets, ...newAssets] : [...newAssets])}
-        />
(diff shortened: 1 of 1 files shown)


apps/cloudinary2/src/locations/Field/index.tsx
```diff
@@ -18,22 +18,26 @@ const Field = () => {
   const sdk = useSDK<FieldAppSDK<AppInstallationParameters>>();
   useAutoResizer({ absoluteElements: true });
 
-  const [assets = [], setAssets] = useFieldValue<CloudinaryAsset[]>(sdk.field.id, sdk.field.locale)
+  const [assets = [], setAssets] = useFieldValue<CloudinaryAsset[]>(sdk.field.id, sdk.field.locale);
+  const assetsIsNotNull = assets !== null;
```
matthew-gordon, line 22: Just curious here, it looks like if useFieldValue returns `null` or `undefined`, the destructuring assignment would ensure that `assets` is set to an empty array so assets should never be null right? It might be redundant 🤔 or am I off base here?

apps/cloudinary2/src/locations/Field/index.tsx
```diff
@@ -18,22 +18,26 @@ const Field = () => {
   const sdk = useSDK<FieldAppSDK<AppInstallationParameters>>();
   useAutoResizer({ absoluteElements: true });
 
-  const [assets = [], setAssets] = useFieldValue<CloudinaryAsset[]>(sdk.field.id, sdk.field.locale)
+  const [assets = [], setAssets] = useFieldValue<CloudinaryAsset[]>(sdk.field.id, sdk.field.locale);
```
jsdalton, line 21: Hey! since we know that `assets` could be null here maybe just make the type argument reflect that `CloudinaryAsset[] | null`

apps/cloudinary2/src/locations/Field/index.tsx
```diff
@@ -18,22 +18,26 @@ const Field = () => {
...
+  const assetsIsNotNull = assets !== null;
 
   const [editingEnabled, setEditingEnabled] = useState(!sdk.field.getIsDisabled());
   useEffect(() => {
     sdk.field.onIsDisabledChanged((disabled) => setEditingEnabled(!disabled));
   }, [sdk.field]);
 
-  const canAddAssets = assets.length < sdk.parameters.installation.maxFiles && editingEnabled;
+  const canAddAssets = assetsIsNotNull && assets.length < sdk.parameters.installation.maxFiles && editingEnabled;
```
jsdalton, line 29: Huh, I'm still not sure what it means for the field to be `null`. So if that's the case this line says you can't add more assets. Is that right?  🤔

matthew-gordon, line 29: per above, might be able to just assert truthy value with `!!assets`

apps/cloudinary2/src/locations/Field/index.tsx
jsdalton: I had a similar line of thinking and questions here, and this also connects to my other comment. If assets is null this later disables the ability to add assets. Not sure if that's what we want? If not, your suggestion sounds like a good one (just fall back to an empty array when null or undefined)

apps/cloudinary2/src/locations/Field/index.tsx
```diff
@@ -18,22 +18,26 @@ const Field = () => {
...
 
   return (
     <>
       <Global styles={globalStyles} />
       <GlobalStyles />
       <Stack spacing="spacingM" flexDirection="column" alignItems="flex-start">
-        {assets.length > 0 && <Thumbnails assets={assets} onChange={setAssets} isDisabled={!editingEnabled} />}
-        <OpenDialogButton isDisabled={!canAddAssets} onNewAssetsAdded={(newAssets) => setAssets([...assets, ...newAssets])} />
+        {assetsIsNotNull && assets.length > 0 && <Thumbnails assets={assets} onChange={setAssets} isDisabled={!editingEnabled} />}
```
matthew-gordon, line 36: same comment here, might be able to just use `!!assets`

jsdalton (approved):
Clean!

PR #1784, opened 2024-05-09
feat: Transifex app remove the default locale restriction

foteinigk:
## Purpose

Remove the check for contentful space default locale match with transifex project source language, since they can be different now

## Testing steps

Try to connect a transifex project with source language different from contentful space's default locale


mkiousi:
Hello from Transifex! Is there an ETA on when the Contentful team will review this PR?

lilbitner (approved):

PR #1828, opened 2024-05-16
feat: adapt-essentials-asset-fields search and bugfixes

zzzarius:
## Purpose

- Add search
- Duplicate inputs bug fix
- Show thumbnails for all selected locales
- Move styles into styles module file
- Show `no assets found` when no assets found
- Read only filename - editing it was causing issues
- Replace deprecated `useCMA` hook with custom one
- Started work on more filtering capabilities
- Other small bugfixes

## Testing steps

- Try searching for media
- Try using/editing using more locales


zzzarius:
cc @whitelisab

PR #1928, opened 2024-05-28
feat: New `amplitude-experiment` app

hthr-lee:
## Purpose

We are introducing a new integration between Amplitude Experiment and Contentful. See [the Amplitude Experiment documentation](https://www.docs.developers.amplitude.com/experiment/integrations/contentful/) for more information. 

## Approach

This queries the Amplitude Experiment Management API to get information about an experiment and prepopulates fields inside Contentful so that the field can match a variant key to a piece of content in Contentful. 

## Testing steps

To test this app, you can run this app locally by running `yarn install && yarn start` and create an app definition pointing to localhost:3001.

## Breaking Changes

This is a new app so there are no breaking changes.

[Amplitude Experiment documentation](https://www.docs.developers.amplitude.com/experiment/integrations/contentful/)


apps/amplitude-experiment/src/hooks/use-entry-card.ts
mgoudy91, line 1: looks like this is just an empty file

apps/amplitude-experiment/src/react-app-env.d.ts
mgoudy91, line 1: same comment, is this needed?

mgoudy91 (changes requested):
Thanks for submitting! This generally looks good, and the only blocking change needed is a LICENSE file, like so: https://github.com/contentful/marketplace-partner-apps/blob/main/apps/flexfields/license.txt

Otherwise, I had a question about some seemingly empty files, but fixes aren't required there necessarily.

mgoudy91 (approved):

mgoudy91:
Looks great @hthr-lee thanks for the updates, I'll get this merged now

PR #2224, opened 2024-06-28
fix: Fix blank editor if Json field is present

vipulchadda:
## Purpose

The FlexFields entry editor was blank when the content type included a JSON field. 

## Approach

Using the JSON field editor explicitly instead of the default field editor for JSON field

## Testing steps

Add JSON field to a content type. Add a rule in FlexFields and verify that the editor displays appropriately.


whitelisab (approved):
Looks good!

PR #2344, opened 2024-07-24
fix: Handled Errors when there is a media without file

vijayaragavan462:
## Purpose

Fixed a bug that was raised by one of the users.

## Approach

When there are some media without a file, we do not need that to show it in the list of existing images.


vijayaragavan462:
Hi team,

This PR includes a critical fix for a high-priority task. Your prompt review would be greatly appreciated, as it’s crucial for us to address this issue as soon as possible.

whitelisab (approved):

PR #2462, opened 2024-08-06
chore: remove vulnerable dependencies for bynder

t-col:
Removes dependencies on unmaintained react-scripts package. Replaces `react-scripts` with `vite`/`vitest`.


david-shibley-contentful (approved):
:shipit:

PR #2468, opened 2024-08-06
chore: remove vulnerable dependencies for shopify

t-col:
Removes dependencies on unmaintained react-scripts and emotion packages. Replaces react-scripts with vite/vitest. Replaces emotion with @emotion/css


david-shibley-contentful (approved):
LGTM

PR #2603, opened 2024-08-23
chore: group dependency upgrades

t-col:
## Purpose

group devDependency and production dependency updates by app


whitelisab (approved):
🌟

Issue #2751, opened 2024-08-30
[FlexFields] DefaultFields for JSON fields using other apps crash the EntryEditor


ajoensson:
**Describe the bug**
When using FlexFields rules in combination with another app that uses JSON fields (such as [the Color Picker app](https://github.com/contentful/apps/tree/master/apps/color-picker)), the entire EntryEditor becomes blank for that content type.

**To Reproduce**
A minimal reproducible example is the following:
1. Install the Color Picker app to your Contentful environment
2. Create a new content type with (at least) the fields below. For our purposes we use the name "Minimal entity" for the content type.
     1. A boolean field. For our purposes we use the name "Show color picker".
     2. A JSON field and choose Color Picker as the appearence/editor. For our purposes we use the name "Custom background color" .
3. Create an entry with the above content type, and set the "Show color picker" field to "Yes".
4. Install the FlexFields app to your Contentful environment (cannot be installed without adding a rule, which is detailed in the next step)
    or if already installed: go to your installed apps and choose "Configure" for FlexFields.
6. Add a new rule as such:
    1. in the "For content type" field, choose the newly created content type (e.g. "Minimal entity")
    2. in the "With field" choose the boolean field (e.g. "Show color picker")
    3. as "Condition", choose "is empty"
    4. under "Hide fields" and "Within content type", choose our content type (e.g. "Minimal entity")
    5. under the "Hide fields", select the Color picker field (e.g. "Custom background color")
    7. Save the app configuration
6. Reload the window to make sure the rule is applied
8. Go to edit your newly created entity. You will be greeted by an empty EntryEditor in the FlexFields tab.

**Expected behavior**
The Entry editor should not be blank. Instead, the boolean field should be able to show/hide the "Custom background color". However, we do not expect the color picker to show up, only the fallback default JSON editor.

**Screenshots**
Screenshot of the blank EntryEditor
<img width="1107" alt="Screenshot 2024-08-30 at 11 05 24" src="https://github.com/user-attachments/assets/480e34fa-a495-49d4-8d12-b83f5dfcf014">
Whereas the standard editor works, but of course doesn't apply the FlexFields rules
<img width="1107" alt="Screenshot 2024-08-30 at 11 08 21" src="https://github.com/user-attachments/assets/7022f69f-db23-4f38-a75a-3e9a89c49ee4">

**Desktop (please complete the following information):**
 - OS: macOS 14.6
 - Browser Google Chrome 
 - Version 127.0.6533.120

[...]

PR #2752, opened 2024-08-30
fix(FlexFields): fallback to the default JSON editor for entry fields that use other apps

ajoensson:
By passing `widgetId` both as a prop and a property of the `control` object resulted in a mismatch between the two when overriding with the default widget ID for the fields that are edited with another app. We now only use the `control` object to choose `widgetId`.

## Purpose

This fixes issue #2751

## Approach

I opted to go with the solution of passing `widgetId` only once via `control`. It would be possible to pass it twice as before, and just make sure we pass the correct value both times, but I would argue that it would open up for similar issues in the future where they get out of sync.

## Testing steps

The reproduction is just like in the issue #2751

Using a JSON field with another app no longer crashes but instead uses the default JSON editor
Screenshot 2024-08-30 at 11 37 04
Screenshot 2024-08-30 at 11 37 06
Of course, the actual app still only works in the normal editor
Screenshot 2024-08-30 at 11 37 10

## Breaking Changes

None!

Issue #2751


david-shibley-contentful (approved):
LGTM, thanks for the contribution!

PR #2820, opened 2024-09-04
fix: adding the public folder back in

david-shibley-contentful:
## Purpose

Somehow the public folder got removed. There are files that depend on that folder such as the raster logo on the config page


whitelisab (commented):
Given that the app is no longer using Next.js, do we need to add the next and vercel svgs? Are they referenced anywhere?

david-shibley-contentful:
Great point, I don't see any references to them. I was just including them cause they were part of the original PR. I will remove them :-)

whitelisab (approved):
🥇

PR #3613, opened 2024-11-15
fix: cloudinary required field

david-shibley-contentful:
## Purpose

[ZEND-5586](https://contentful.atlassian.net/browse/ZEND-5586)

## Deployment

[ZEND-5586]: https://contentful.atlassian.net/browse/ZEND-5586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sarahlessner (approved):

apps/cloudinary2/src/locations/Field/index.tsx
```diff
@@ -33,7 +33,13 @@ const Field = () => {
       <Global styles={globalStyles} />
       <GlobalStyles />
       <Stack spacing="spacingM" flexDirection="column" alignItems="flex-start">
-        {assets.length > 0 && <Thumbnails assets={assets} onChange={setAssets} isDisabled={!editingEnabled} />}
+        {assets.length > 0 && <Thumbnails assets={assets} onChange={(asset) => {
```
whitelisab, line 36: I was looking at the behavior of the Bynder app, and I think that it would be better for us to follow the pattern of that app, where when there is not selected asset, the field value is removed entirely. It does this by using the `sdk.field.removeValue` function. Here is an example of how that is implemented in the dam-app-base, which the Bynder app uses: https://github.com/contentful/apps/blob/master/packages/dam-app-base/src/Editor/Field.tsx#L86-L93. Might require a bit more work, but it would follow the pattern of when a field is empty, it is entirely removed from the field object

whitelisab (commented):
Added a comment about the approach for you to consider!

apps/cloudinary2/src/locations/Field/index.tsx
david-shibley-contentful: Can do, thanks for looking into that!

PR #4175, opened 2025-01-23
chore: app review workflow works for forked branches[INTEG-2165]

sarahlessner:

### DIFF
--- a/.github/workflows/new-app-review.yml
+++ b/.github/workflows/new-app-review.yml
@@ -15,8 +15,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
         with:
+          # Fetch all history for all branches and tags
+          fetch-depth: 0
+          # Ensure the workflow can access the forked repository's contents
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "@actions/github": "^6.0.0",
     "@types/node": "^22.3.0",
     "husky": "^8.0.0",
-    "lint-staged": "^13.2.1",
+    "lint-staged": "^15.4.2",
```
sarahlessner, line 14: Just a dependency update that came up running npm audit that didn't have any breaking changes to worry about with the fix.

.github/workflows/new-app-review.yml
```diff
@@ -15,8 +15,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
         with:
+          # Fetch all history for all branches and tags
```
sarahlessner, line 19: This is a solution I found to the issue, but I can't really think of a way to test this out without having this workflows file merged in first. If I made this change on a forked repo branch I don't feel confident that it would work reliably.

mgoudy91 (approved):

PR #4289, opened 2025-02-06
fix: sfmc studio app remove fonts [INTEG-2359]

david-shibley-contentful:


david-shibley-contentful:
These absolute paths will need to be changed to relative paths

sarahlessner (approved):
Nice work on this one, it'll be great to finally get it released!

PR #4491, opened 2025-02-27
feat: adds labels to dependabot prs for contentful owned apps [INTEG-2430]

sarahlessner:
Some apps in this repo are supported by Contentful. When dependabot PRs are open it would be nice to see at a quick glance whether or not the PR is something we can approve and update that dependency ourselves versus getting approval from the app developer. I added a label to the apps we own that have a dependabot configuration.


mgoudy91 (approved):
love this

PR #4529, opened 2025-03-04
chore: update deps in contentful owned apps [INTEG-2453]

sarahlessner:
There are some critical/high vulnerabilities in the Vite, Vitest, Happy-dom and esbuild packages so I am updating those for the apps we own in this repo.

### DIFF
--- a/apps/bynder/package.json
+++ b/apps/bynder/package.json
@@ -5,7 +5,7 @@
   "devDependencies": {
     "@contentful/app-scripts": "^2.1.0",
     "@vitejs/plugin-react": "^4.3.4",
-    "happy-dom": "^15.11.7",
+    "happy-dom": "^17.1.9",
     "vite": "^6.2.0",
     "vitest": "^3.0.7"
   },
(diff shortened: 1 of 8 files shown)


apps/bynder/package.json
```diff
@@ -3,16 +3,16 @@
...
-    "vite": "^5.4.11",
-    "vitest": "^2.1.8"
+    "vite": "^6.2.0",
+    "vitest": "^3.0.7"
   },
   "dependencies": {
     "@contentful/dam-app-base": "^3.0.9",
-    "react": "18.3.1",
-    "react-dom": "18.3.1"
+    "react": "18.2.0",
```
sarahlessner, line 14: when I tried to `npm i` from the bynder root, I was getting a dependency conflict due to the react version. seems like dependabot updated it _6 months_ ago. I don't quite understand how it was passing.  I looked at our `dam-app-base` package and it's using React 18.2 so I put this back to resolve.

david-shibley-contentful (approved):

PR #4603, opened 2025-03-07
feat: lottie viewer app

shanonplace:
This Contentful App Framework app augments the JSON field editor to also include a Lottie player, allowing the user to see the animation in action. 🎥

**Key Features:**

- Runs as a field editor, incorporating a Lottie Preview component next to the JSON editor field
- Allows the editor to see the rendered Lottie animation inside the Contentful web app


shanonplace:
Need to fix some issues in the build locally first

PR #4746, opened 2025-03-28
fix: Italics in content causes a validation error when importing [INTEG-2515]

CWDN:
## Purpose

This is fixing a bug with the Content Workflow Bynder integration where if the user had italic content in their fields in Content Workflow the import would throw a validation error.

## Approach

We used to have an override for the `em` tag to create a nested paragraph. This isn't compatible with the Contentful API so instead dropping our custom handler and bumping the `contentful-rich-text-html-parser` to the latest release which has `em` support already.

## Testing steps

You would need a Content Workflow account but simply put italic content into a Content Workflow item field, then setup the template mapping between the Content Workflow and Contentful. Then when you try to import the content it would break.

## Breaking Changes

Not as far as I can tell, tested everything on my local to ensure it all worked.

Sorry all internal tickets that have been raised by a customer.

### DIFF
--- a/apps/bynder-content-workflow/package-lock.json
+++ b/apps/bynder-content-workflow/package-lock.json
@@ -1,26 +1,27 @@
 {
   "name": "bynder-content-workflow",
-  "version": "1.0.8",
+  "version": "1.0.9",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "bynder-content-workflow",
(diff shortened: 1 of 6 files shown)


primeinteger (approved):

HughbertD:
@primeinteger Sorry to bother, but is it possible for a quick re-review and deploy? We needed to push a change just now

PR #4761, opened 2025-03-31
translationstudio app added

translationstudio:
## Purpose

This is a PR for our marketplace submission as requested by Matt M. 

## Approach

PR requested during marketplace submission process

## Testing steps

As I understand, you will request a demo anyway, so no need for testing here

## Breaking Changes

Nope, it is a separate app

App added due to marketplace submission

## Deployment

.


sam-is-content:
Hello! I've been testing out the app with your recent changes - the changes look great but I'm seeing the occasional problem with the translation returning and populating the field. Most of the time, when I enter a word then press translate, I do not see the translation come back no matter how long I wait. However, if I press the translate button a second time, the translation comes back quickly and populates the field.

translationstudio:
Hi @sam-is-content ,

I quickly checked the issue and would like to know which fields you tried to translate. 

In a quick test, I found that rich text fields are not updated automatically in the background at all times. Plain text fields do, but sometimes I have to refresh the browser page to force a rich text reload. Can you provide additional details about your entry?

sam-is-content:
@translationstudio here is a video of what I am seeing - this was for short text. Let me know if you have any trouble viewing the video. https://www.loom.com/share/6ba8ad0fd90a4b66805473f4e6381442?sid=ffc0b7e2-e2e9-4791-8f17-f3464e7973f7

translationstudio:
Hi @sam-is-content 

I can reproduce this behaviour and it seems that there is an issue with the Contentful API.  I have implemented a workaround on our end (not part of this PR), so that case should be fine and your testing succeed.

The issue is described here: https://github.com/contentful/contentful-management.js/issues/2644

sam-is-content:
> Hi @sam-is-content
> 
> I can reproduce this behaviour and it seems that there is an issue with the Contentful API. I have implemented a workaround on our end (not part of this PR), so that case should be fine and your testing succeed.
> 
> The issue is described here: [contentful/contentful-management.js#2644](https://github.com/contentful/contentful-management.js/issues/2644)

Just to be sure that I am understanding you, should the code at https://contentful.translationstudio.tech/ include the workaround or no?

translationstudio:
Hi @sam-is-content correct, the fix is not related to the core running at contentful.translationstudio.tech - it is at a nother component that this app talks to when it comes to perform export and import.

sam-is-content:
@translationstudio hello! The app looks good from the provided URL but I was unable to build and host the app via Contentful App Hosting. For an app to be part of the Contentful Marketplace it must be hosted by Contentful App Hosting and Contentful App Hosting can only host static websites. The routes use a server and thus the site is not static. You can try refactoring the server code to use [App Functions](https://www.contentful.com/developers/docs/extensibility/app-framework/working-with-functions/) instead. Please let me know how this goes, if the documentation is helpful, and if you find limitations with the App Framework.

apps/translationstudio/components/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,131 @@
...
+						className={css({ width: "100%", marginRight: "20px" })}
+						placeholder="translationstudio License"
+						name="translationStudioKey"
+						id="translationStudioKey"
+						rows={2}
+						value={parameters.translationStudioKey || ""}
+						onChange={onKeyChange}
+						isRequired
+					/>
+					<Button variant="positive" isDisabled={!parameters.translationStudioKey || validKey} onClick={onKeyCheck}>Validate License</Button>
```
primeinteger, line 119: I'm hitting a CORS error when trying to Validate License on the Configuration page.

Screenshot 2025-06-27 at 8 52 37 AM

This error happens when running the app on `http://localhost:3000` and when running the app in a bundled deployment on `https://(.*).ctfcloud.net` (which is how the app will run when it is shared publicly on the Contentful Marketplace)

> Access to fetch at 'https://contentful.translationstudio.tech/translationstudio/validate' from origin 'https://d1ed428b-732b-4270-8008-7bfb73286a3c.ctfcloud.net' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The 'Access-Control-Allow-Origin' header has a value 'https://cms-contentful-service-965a75018a3f.herokuapp.com' that is not equal to the supplied origin. Have the server send the header with a valid value.

Could you please look into this? You may need to configure your License Validation server to enable CORS for requests sent from both `http://localhost:3000` and `https://(.*).ctfcloud.net` (where `(.*)` represents a wildcard to match any subdomain since each version deployed can generate a different subdomain).

apps/translationstudio/next.config.js
```diff
@@ -0,0 +1,6 @@
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  reactStrictMode: true,
+};
```
primeinteger, line 4: Could you please update the Next Config to enable these additional settings?

These settings will adjust `npm run build` to generate the build output files in the `./out` directory as static html (index.html, 404.html, etc). This is necessary so that we can upload the bundled app to our deployment server so the app can be served from `https://(.*).ctfcloud.net`

```suggestion
const nextConfig = {
  reactStrictMode: true,
  basePath: '',
  assetPrefix: './',
  output: 'export',
  trailingSlash: true,
  images: {
    unoptimized: true,
  },
};
```

primeinteger (dismissed):
Hi @translationstudio, we appreciate the updates you made to refactor the app.

I continued our app review process and I'm running into issues with the Validate License encountering CORS errors on the Configuration screen, and a couple other small items that I commented on below.

Could you please look into this and let us know when you are ready for us to review again? Thank you!

translationstudio:
Hi @primeinteger 

updates were made and PR updated. Review can proceed.

sarahlessner:
Hi @translationstudio I've taken a look at the recent changes and it's looking good overall. I was able to upload your app bundle and load the app successfully.

Only issue I notice is our circleci test job keeps getting hung up on running the translationstuido tests because of the `--watch` flag - can we please remove that from your package.json test command? it can be simply `test:jest` and if you need the `--watch` command for local development, would recommend adding in a different command for that, as our ci job relies on the test command.

translationstudio:
Hi @sarahlessner ,

Of course. We do not need that specifically, so I changed it as requested and updated the PR accordingly.

Best,
Phillip.

translationstudio:
Hi @sarahlessner 

for now, I have just updated the PR with the flag --passWithNoTests  

We will update it with the next feature update.

Best,
Phillip.

sarahlessner (approved):

PR #4784, opened 2025-04-03
feat: [INTEG-2513]

tylerwashington888:
## Purpose

add validation to check for boilerplate code


david-shibley-contentful (approved):

.github/workflows/new-app-review/validators/boilerplate.ts
```diff
@@ -0,0 +1,32 @@
...
+    .some((file) => {
+
+      if (file.patch) {
+        const paragraphComponentRegex = /<Paragraph[^>]*>([\s\S]*?)<\/Paragraph>/i;
+        const paragraphComponent = file.patch.match(paragraphComponentRegex);
+
+        if (paragraphComponent) {
+          const innerContent = paragraphComponent[1];
+          const helloRegex = /Hello/i;
+          return helloRegex.test(innerContent);
```
primeinteger, line 16: There might a small potential this could cause a false positive, for example if the partner's app included `<Paragraph>Hello {user}</Paragraph>`

The regex could be expanded to avoid that, although this is probably a small chance of ever having a false positive

```ts
const helloRegex = /Hello \w.+ Component/i;
```

primeinteger, line 16: I noticed the ConfigScreen boilerplate component has some slightly different wording, instead of "Hello {some} Component" it says "Welcome to your contentful app. This is your config page." [shown here](https://github.com/contentful/create-contentful-app-examples/blob/main/v2/templates/alpha-typescript/src/locations/ConfigScreen.tsx#L63)

Should we add another regex check here to look for this string too?

.github/workflows/new-app-review/validators/boilerplate.ts
```diff
@@ -0,0 +1,32 @@
...
+        }
+        return false
```
primeinteger, line 18: Minor nit, but this `return false` isn't totally necessary here since the end of the function will already `return false`

primeinteger (commented):
I didn't review this fast enough. Looks like it was already merged but here's the feedback I wanted to share

PR #4785, opened 2025-04-03
feat: [INTEG-2513] 

tylerwashington888:
## Purpose

update boilerplate validation logic


.github/workflows/new-app-review/validators/boilerplate.ts
```diff
@@ -5,25 +5,29 @@ export const validate = async (_options: ValidatorOptions, newAppDir: string, fi
...
       }
       return false;
     });
 
   const result = hasInvalidBoilerPlate;
   const message = result
-    ? 'Boilerplate check passed'
-    : 'Boilerplate check failed: Found components in the locations folder with invalid boilerplate code. Please remove these components and their respective test files and try again.';
+    ? 'All location components are valid.'
+    : 'Unused locations found: Found components in the locations folder that are unchanged from the code generated by create-contentful-app. Please remove any location component and test files for any locations that are not rendered by your app"';
```
primeinteger, line 30: 👀 spotted an extra `"` at the end of the error message string

primeinteger (approved):
Looks great!  One comment about an extra character, but approving either way since that's so minor

.github/workflows/new-app-review/validators/boilerplate.ts
tylerwashington888: great catch prime integer. just fixed this

PR #4811, opened 2025-04-08
[INTEG-2552] feat: add lottie-viewer initial setup

ryunsong-contentful:
## Purpose
Initialize the project 

App is installed here https://app.contentful.com/spaces/q72hxe9cm2jc/environments/master/entries/15ljN0kUtjuWljTSrMfsQe?isNewEntry=true&focusedField=animate


apps/lottie-viewer/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "lottie-viewer",
+  "version": "1.0.0",
```
sarahlessner, line 3: could having the version already at 1 cause some issues? I know our app review guidelines say the version has to be less than 1. If it's a later problem though then ignore this!

release-please-config.json
```diff
@@ -14,6 +14,7 @@
 		"apps/imageHotspotCreator": {},
 		"apps/intershop": {},
 		"apps/livestory": {},
+		"apps/lottie-viewer": {},
```
sarahlessner, line 17: do we want to just remove this from here until we're ready to release? Maybe it's harmless

sarahlessner (approved):
Couple comments but no concerns about merging while in progress

apps/lottie-viewer/package.json
ryunsong-contentful: The other apps have 1.x.x on their versions. Does the team have a convention on this - happy to follow it? If this causes an error I would like to trigger it and investigate the cause since that in it of itself seems a little odd.

release-please-config.json
ryunsong-contentful: Hmm release-please is just a PR and changelog helper. It doesn't actually deploy anything but if this is a convention the team follows then we can do it when we're ready to go to production. But release-please itself doesn't affect the deploy or build anything. It's purely a PR and changelog tool

apps/lottie-viewer/package.json
sarahlessner: The process works that the `release-please` manifest will automatically set the app version to `1.0.0` at the time we go through the app release flow, and it'll handle incrementing the version when subsequent changes are made to the app.

sarahlessner: It also creates/updates a changelog.
https://contentful.atlassian.net/wiki/spaces/ECO/pages/4334125728/Marketplace+Partner+Apps+-+App+Review+Process#To-accept-PRs-before-app-is-ready-to-be-published-to-the-Marketplace-(in-development)

release-please-config.json
sarahlessner: My thinking was if we set the version to 1 now, then we keep making development changes, the changelog will be generated and the version/changelog will keep getting updated. I would think we would want to wait until our development is done and the initial release is ready for it to be considered version 1 and subsequent version updates/changelog items would be generated from that point forward.

apps/lottie-viewer/package.json
ryunsong-contentful: Ahh very interesting - great call out. I set it to 0.0.1

release-please-config.json
ryunsong-contentful: Hmm generally I think it's okay to start a release with 1.x.x since release versions are for us developers and not the customer but lets go with your idea and let me create a ticket to update this otherwise I'm definitely going to forget this small detail.

sarahlessner: That's a fair point - this is a little out of the normal flow so it's a good time to poke at this.

We have a ticket for releasing the app I think, and when that time comes we can follow the standard process which details adding things to release please, adding the dependabot config, etc.

PR #4856, opened 2025-04-15
feat: add modals to Lottie Preview Viewer for json and animation [INTEG-2552]

ryunsong-contentful:
## Purpose

https://github.com/user-attachments/assets/de84ca95-37ab-4e98-9f4d-b0dc78ef1c5f

### DIFF
--- a/apps/lottie-viewer/src/components/modals/JsonEditorModal.tsx
+++ b/apps/lottie-viewer/src/components/modals/JsonEditorModal.tsx
@@ -60,7 +60,6 @@ export default function JsonEditorModal(props: Props) {
         if (model) {
           model.setValue(initial);
           model.pushStackElement(); // Mark as baseline
-          model.clearUndoStack();   // Reset undo/redo
         }
         updateUndoRedo();
       }, 0);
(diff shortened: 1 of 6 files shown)


apps/lottie-viewer/src/components/modals/JsonEditorModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+import { css } from 'emotion';
+import Editor from '@monaco-editor/react';
+
+type Props = {
+  showJsonModal: boolean;
+  onShowJsonModalChange: (show: boolean) => void;
+  onSave: (value: string) => void;
+  onEditorWillMount: (monaco: any) => void;
+  updateUndoRedoState: () => void;
+  lottieJson: any;
```
sarahlessner, line 15: Is there a reason for the `any` type? I see a few in this file.

ryunsong-contentful: I don't know how to define a Lottie JSON type. Maybe I can look it up

ryunsong-contentful: Nvm I was able to figure it out and define a lottie json type. Good callout

ryunsong-contentful:
> Big nitpick you're going to love but you've got "model" instead of "modal" in a lot of places

Hmm, let me rename it from `model` to `editorModel` to make the variables more distinct from each other. Your comment made me think that the code in this PR can be improved in clarity with respect to naming. So I'll make that change

PR #4864, opened 2025-04-15
fix: loading lottie preview json loads correctly [INTEG-2576]

ryunsong-contentful:
## Purpose
Basically separate the Field component that will load the data from the sdk field and the actual LottieJsonField component so that everything renders in order.

https://github.com/user-attachments/assets/03dc71eb-edff-4d0f-bc1a-deae09e10b29


sarahlessner (approved):
Looks good, nice work. Lots of refactoring to fix the bug!

PR #4895, opened 2025-04-18
feat: adding hugging face, updating config page, updating .gitignore [INTEG-2584]

david-shibley-contentful:
## Purpose

[INTEG-2584](https://contentful.atlassian.net/browse/INTEG-2584)

## Approach

Screenshot 2025-04-18 at 2 45 52 PM

## Deployment

[INTEG-2584]: https://contentful.atlassian.net/browse/INTEG-2584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tylerwashington888 (approved):
LGTM

apps/ctfl-hugging-face-image-gen/README.md
primeinteger, line 1: What determines when apps are added to the `apps` repo instead of the `marketplace-partner-apps` repo?  It seems like more of our internally owned apps are in the other repo

apps/ctfl-hugging-face-image-gen/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "prompt-image-hf",
```
primeinteger, line 2: I wonder if the `ctfl-` should be dropped from the directory for this app? If the app is called "Hugging Face" in the Marketplace, having `ctfl-` in the path would probably make it harder for me to find this apps source code is.

Also wonder if this name in package.json should be updated to match the folder.  Not sure if it matters but then it would have a matching name when running build commands and such.

apps/ctfl-hugging-face-image-gen/README.md
david-shibley-contentful: That's a good question, I guess I just made an assumption this went here but I could be wrong

apps/ctfl-hugging-face-image-gen/package.json
david-shibley-contentful: Yeah I like it!

PR #4913, opened 2025-04-23
feat: [INTEG-2557] Lottie Viewer config page

sarahlessner:
## Purpose

Lottie Viewer can be set as the "appearance" for a JSON field type in the content model. We allow the users to set an applicable field if one exists in their content model from the configuration page:

image

If there are no matching Content Types:

image

Dropdown open state:

image

Filtered:

image

## Approach

Allows users to set from the install page instead of from their content type page, if they have a relevant content type to use. 

## Testing steps

Head to the config page for a space with and without JSON field types in its content model.

Should see the empty state if you don't

Should be able to check/uncheck and verify via the content model that the correct "appearance" is set for the field


apps/lottie-viewer/src/configUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+
+  return jsonFields.sort((a, b) => {
+    const typeCompare = a.contentTypeName.localeCompare(b.contentTypeName);
+    return typeCompare !== 0 ? typeCompare : a.fieldName.localeCompare(b.fieldName);
+  });
+}
+
+export function groupFieldsByContentType(fields: JsonField[]): Record<string, JsonField[]> {
```
sarahlessner, line 48: Updating the interface for the content types on save involves making updates to the underlying fields on a per-content type update basis, so this batches them by content type for the case where more than one JSON field in a single content type had this treatment applied

apps/lottie-viewer/src/configUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+  return fields.reduce((acc, field) => {
+    if (!acc[field.contentTypeId]) acc[field.contentTypeId] = [];
+    acc[field.contentTypeId].push(field);
+    return acc;
+  }, {} as Record<string, JsonField[]>);
+}
+
+export function buildEditorInterfaceControls(allFields: JsonField[], existingControls: Control[] = [], appId: string): Control[] {
```
sarahlessner, line 56: the editorInterface is the format the cma is expecting this to all come back in, where the widgetId and widgetNamespace are ultimately what control the "appearance" of the field being either its default Object or a given app.

apps/lottie-viewer/src/hooks/useJsonFieldsState.ts
```diff
@@ -0,0 +1,39 @@
+import { JsonField } from '@src/configUtils';
+import { useRef, useState } from 'react';
+
+type JsonFieldUpdate = Partial<Pick<JsonField, 'isEnabled'>>;
+
+export function useJsonFieldsState() {
```
sarahlessner, line 6: There was a lot of state/ref logic related to managing the json fields - initializing them, updating them, etc. and I couldn't just have local state to handle rendering them as items in the dropdown/selected items. I also needed to keep track of them for ultimately making the updates on save. The way onConfigure works, the state winds up getting cleared when saving so the ref is for keeping track of the changes so we can batch them and send them on save.

apps/lottie-viewer/src/locations/ConfigScreen.tsx
```diff
@@ -0,0 +1,147 @@
...
+    installTriggeredRef.current = true;
+    return { parameters };
+  }, [parameters]);
+
+  useEffect(() => {
+    sdk.app.onConfigure(onConfigure);
+  }, [sdk, onConfigure]);
+
+  useEffect(() => {
+    sdk.app.onConfigurationCompleted(async () => {
```
sarahlessner, line 40: I had to add this because onConfigure kept overwriting the changes I'd make to update `editorInterface` so I have to watch for when the other  configuration steps complete and then send the updates. I have a ref to know whether this effect ran because the install was triggered. But this is where we're actually sending those batch updates through the CMA and then resetting the state.

mgoudy91 (approved):
I think this makes sense. Honestly I was pretty surprised at how complex of a solution this is for just saving content types, but I also know your approach of handling editor interfaces in this way is correct, so I don't really have any feedback there.

This appears to work as expected for me and think it's pretty clean UX, improved over the more common multiselect we have in other apps.

Nice work!

sarahlessner:
> Honestly I was pretty surprised at how complex of a solution this is for just saving content types

@mgoudy91 that makes two of us 😬

ryunsong-contentful:
Great work, couldn't find any bugs and tested in all kinds of different situations. Very nicely done.

One small comment which is totally up to you if you want to push that little 0.1% further.
1. 
Screenshot 2025-04-23 at 3 51 21 PM

```There are no JSON object field types to select to use with Lottie Preview. Once you have added one to a content type, it will appear here.```

Could be helpful to just define a little more what `here` means - so something like `will enable the dropdown with a list of Content types with a JSON object field type` for example.

ryunsong-contentful (approved):

sarahlessner:
I think that's a good improvement, I don't even know if I really read it versus just blindly copy/pasting from Figma 😆

PR #5007, opened 2025-05-16
feat: post release cleanup [INTEG-2586]

mgoudy91:
## Purpose

- Use space default locale instead of en-us

- Use better default params for config model options

- fix the dang tests

### DIFF
--- a/apps/hugging-face/src/locations/ConfigScreen.tsx
+++ b/apps/hugging-face/src/locations/ConfigScreen.tsx
@@ -9,8 +9,8 @@ import { INFERENCE_PROVIDERS } from '@huggingface/inference';
 const ConfigScreen = () => {
   const [parameters, setParameters] = useState<AppInstallationParameters>({
     huggingfaceApiKey: '',
-    textModelId: 'meta-llama/Llama-3.2-3B-Instruct',
-    textModelInferenceProvider: 'hf-inference',
+    textModelId: 'Qwen/Qwen2.5-VL-7B-Instruct',
+    textModelInferenceProvider: 'hyperbolic',
     imageModelId: 'black-forest-labs/FLUX.1-dev',
     imageModelInferenceProvider: 'hf-inference',
   });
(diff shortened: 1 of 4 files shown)


apps/hugging-face/src/locations/ConfigScreen.tsx
```diff
@@ -9,8 +9,8 @@ import { INFERENCE_PROVIDERS } from '@huggingface/inference';
 const ConfigScreen = () => {
   const [parameters, setParameters] = useState<AppInstallationParameters>({
     huggingfaceApiKey: '',
-    textModelId: 'meta-llama/Llama-3.2-3B-Instruct',
-    textModelInferenceProvider: 'hf-inference',
+    textModelId: 'Qwen/Qwen2.5-VL-7B-Instruct',
+    textModelInferenceProvider: 'hyperbolic',
     imageModelId: 'black-forest-labs/FLUX.1-dev',
```
sam-is-content, line 14: i think we should also find an image model that doesn't first require a consent too

apps/hugging-face/src/services/uploadAsset.ts
```diff
@@ -18,19 +23,20 @@ const createAsset = async ({
   const file = new File([blob], fileName, { type: 'image/png' });
   const buffer = await file.arrayBuffer();
   const upload = await sdk.cma.upload.create({ spaceId: sdk.ids.space }, { file: buffer });
+  const defaultLocale = await getDefaultLocale(sdk);
```
sam-is-content, line 26: if the defaultLocale ends up (somehow) returning `undefined` will the code below still work>

primeinteger (approved):
Looking good 👏

apps/hugging-face/src/locations/ConfigScreen.tsx
mgoudy91: oh I thought that one did. ill look around

apps/hugging-face/src/services/uploadAsset.ts
mgoudy91: not possible (in terms of business logic) currently. I can add a super careful backup though

Issue #5136, opened 2025-06-06
Shopify App - Search by SKU not working anymore


chrisvltn:
**Describe the bug**

Since few days ago, the search by SKU doesn't work anymore in the product variants selector app.

Search by title still works.

**To Reproduce**
1. Open the product variant selector
2. Search for a variant SKU

**Expected behavior**
The desired variant to appear

**Screenshots**

Before:

![Image](https://github.com/user-attachments/assets/69f7cba2-8449-4320-97c8-1b42c74ea1f7)

After: 

![Image](https://github.com/user-attachments/assets/326a76f7-9153-4e45-a0d0-9e3ce91c2015)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 136.0.7103.116

**Additional context**

I see this PR (https://github.com/contentful/marketplace-partner-apps/pull/5130) did a big refactor, and the query by variant SKU changed from `variants:['sku:${search}']` to `variants.sku:*${search}*`.

https://github.com/contentful/marketplace-partner-apps/pull/5130/files#diff-dd65a7241c6ffe170d588e82ab7fc5e3ec699f0389b75a52ef10e6daa3dd9b61R10

I tested the Storefront API with the previous query format and it worked for our shop. Maybe this could be a potential fix?

chrisvltn:
This seems to also happen for collections. Before it was possible to search by handle, ID, title, etc.

Now it is only possible to search by title.

https://github.com/contentful/marketplace-partner-apps/pull/5130/files#diff-c22139f43316d9a72d6574108b48a4d9223afe510dc6875f6ee594a177214ba9R10

primeinteger:
Thank you for reporting this issue @chrisvltn. We'll look into it

mgoudy91:
Hey @chrisvltn , like @primeinteger said we appreciate the report!

We've deployed a fix that we believe should solcv

chrisvltn:
Nice! I can confirm this works for us now 👍 Thanks a lot for the super fast reaction! We really appreciate that 🎉

PR #5137, opened 2025-06-06
fix: sku searching fixes and wildcard fixes

mgoudy91:
## Purpose

Fixes:  https://github.com/contentful/marketplace-partner-apps/issues/5136

- SKU search was broken with last migration to the storefront API

https://www.loom.com/share/ce985fffff06447f9c7582bbed8f0229?sid=0c8e6667-eec4-4e78-86fd-4e81afa901e0


sarahlessner (approved):
More work to fix that than I realized that was quick! Let's hope this does it 🤞

PR #5139, opened 2025-06-06
fix: performance issues on lottie viewer config page [ZEND-6496]

sarahlessner:
## Purpose

A customer reported being unable to install the Lottie Preview app. Impersonating them, I see they have a lot of Content Types. I reviewed the logic for the config page knowing how it fetches many content types and deals with their editor interfaces to apply the Lottie app to JSON fields within Content Types. 

There were a few critical issues that I believe will resolve this for the customer.

### DIFF
--- a/apps/lottie-preview/src/locations/ConfigScreen.tsx
+++ b/apps/lottie-preview/src/locations/ConfigScreen.tsx
@@ -24,8 +24,13 @@ const ConfigScreen = () => {
         isChecked: field.isEnabled,
         contentTypeId: field.contentTypeId,
       })),
-    [jsonFields, inputValue]
+    [jsonFields]
   );
+
+  const filteredItems = useMemo(() => {
+    if (!inputValue) return items;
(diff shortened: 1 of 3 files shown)


apps/lottie-preview/src/locations/ConfigScreen.tsx
```diff
@@ -42,18 +47,21 @@ const ConfigScreen = () => {
 
       const fieldsByContentType = groupFieldsByContentType(jsonFieldsRef.current);
 
-      for (const [contentTypeId, allFields] of Object.entries(fieldsByContentType)) {
+      // Process all content types in parallel
+      const updatePromises = Object.entries(fieldsByContentType).map(async ([contentTypeId, allFields]) => {
```
sarahlessner, line 51: Similar performance improvement as aggregating the JSON fields for when we update the content types on config. This isn't as likely to be a problem since it'd be rare for someone to make a lot of updates in one go but why not make it better

PR #5146, opened 2025-06-10
fix: lottie preview config page fails to load with many content types [ZEND-6496]

sarahlessner:
## Purpose

A customer reported that the Lottie config page wouldn't load for them so they couldn't install the app. Impersonating their space I saw they had a lot of content types, and optimized how we were fetching the editor interfaces. This still didn't resolve the issue, so I'm going a bit more excessive here.

We have a loading state for batching lots of editor interface fetching now, this shouldn't happen in ordinary use cases:

Screenshot 2025-06-10 at 9 56 05 AM

but in the absence of an SDK built in method to handle this efficiently, this is the best we can do for now.

I invited everyone to my test space with lots of content types: https://app.contentful.com/spaces/nmiq2054z3yp/apps/2HjkWl9ilqInZlWjfGhWh5

Note that this space has 303 content types with JSON fields. I suspect that would be pretty unusual without knowing for certain. The customer who reported this has about 300, but I spot checked several and only a couple had JSON fields, and we wouldn't fetch all the editor interfaces for them in that case. 

can also try installing this staging app in a normal space you use


apps/lottie-preview/src/configUtils.ts
```diff
@@ -4,6 +4,16 @@ const JsonFieldType = 'Object';
 const AppWidgetNamespace = 'app';
 const DefaultWidgetId = 'objectEditor';
 
+// Batch configuration for API requests
```
sarahlessner, line 7: some of this is just guessing/was testing with using the AI and it worked for my 300 content types with JSON fields. any input here is definitely welcome

apps/lottie-preview/src/configUtils.ts
```diff
@@ -19,59 +29,197 @@ export interface JsonField {
...
+  appDefinitionId: string,
+  options?: { limit?: number; offset?: number },
+  onProgress?: (processed: number, total: number) => void
+): Promise<JsonFieldsResult> {
+  const contentTypes = await cma.contentType.getMany({
+    query: { limit: 1000 },
+  });
+
+  // Filter to only content types that have JSON Object fields
+  const contentTypesWithJson = contentTypes.items
```
sarahlessner, line 113: I think this change here might cause things to be a lot more performant in most cases alone, only fetching editor interfaces for content types that actually have the field in question.

apps/lottie-preview/src/locations/ConfigScreen.tsx
```diff
@@ -113,40 +246,62 @@ const ConfigScreen = () => {
...
+          </>
         ) : (
-          <Flex className={styles.pillsRow}>
-            {items
-              .filter((item) => item.isChecked)
-              .map((item) => (
-                <Pill testId={`pill-${item.id}`} key={item.name} label={item.name} onClose={() => handleSelectItem(item)} />
-              ))}
-          </Flex>
+          <>
```
sarahlessner, line 266: nothing really changed here this is just the default load the Autocomplete UI

apps/lottie-preview/src/configUtils.ts
sarahlessner: Looks like for CMA calls it's 10 per second so batches of 10 does seem good

apps/lottie-preview/src/configUtils.ts
```diff
@@ -4,6 +4,16 @@ const JsonFieldType = 'Object';
 const AppWidgetNamespace = 'app';
 const DefaultWidgetId = 'objectEditor';
 
+// Batch configuration for API requests
+const BATCH_SIZE = 10; // Process editor interfaces in batches of 10
+const RETRY_DELAY = 500; // Base delay for retries (ms)
+const MAX_RETRIES = 3; // More retries with better backoff
+const OVERALL_TIMEOUT = 120000; // 60 second timeout for large content models
```
primeinteger, line 11: There's a small discrepancy between the value and comment for overall timeout

apps/lottie-preview/src/configUtils.ts
```diff
@@ -19,59 +29,197 @@ export interface JsonField {
...
+      processedContentTypes: 0,
+      hasMore: false,
+    };
+  }
+
+  const offset = options?.offset || 0;
+  const limit = options?.limit || INITIAL_BATCH_SIZE;
+  const totalWithJsonFields = contentTypesWithJson.length;
+
+  const startIndex = offset;
```
primeinteger, line 133: [nit pick] Could use `offset` instead of creating this alias variable `startIndex` - or rename `offset` to `startIndex` if it's better for readability

apps/lottie-preview/src/configUtils.ts
```diff
@@ -19,59 +29,197 @@ export interface JsonField {
...
+    for (const contentType of contentTypes.items) {
+      const hasJsonFields = contentType.fields.some((f) => f.type === JsonFieldType);
+      if (hasJsonFields) {
+        count++;
+      }
+    }
+
+    return count;
+  } catch (error: any) {
+    console.error('getContentTypesWithJsonFieldsCount failed:', error.message);
```
primeinteger, line 77: Are you sure this error will always have a `message` property? Some error objects don't and it could throw an error itself if it catches an error that doesn't have a `message` prop.  It might be safer to log the full `error` object in the second param instead of `error.message`

apps/lottie-preview/src/configUtils.ts
```diff
@@ -19,59 +29,197 @@ export interface JsonField {
...
+  cma: ConfigAppSDK['cma'],
+  appDefinitionId: string,
+  options?: { limit?: number; offset?: number },
+  onProgress?: (processed: number, total: number) => void
+): Promise<JsonFieldsResult> {
+  try {
+    const result = await withTimeout(getJsonFieldsInternal(cma, appDefinitionId, options, onProgress), OVERALL_TIMEOUT);
+    return result;
+  } catch (error: any) {
+    console.error('getJsonFields failed:', error.message);
```
primeinteger, line 92: Similar concern here about `error.message` possibly throwing its own error if the caught error doesn't have a `message` property

apps/lottie-preview/src/locations/ConfigScreen.tsx
```diff
@@ -1,41 +1,63 @@
...
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { styles } from './ConfigScreen.styles';
-import { ExternalLinkIcon } from '@contentful/f36-icons';
+import { ClockIcon, ExternalLinkIcon } from '@contentful/f36-icons';
 import { useJsonFieldsState } from '@src/hooks/useJsonFieldsState';
-import { buildEditorInterfaceControls, getJsonFields, groupFieldsByContentType } from '@src/configUtils';
+import { getJsonFields, getContentTypesWithJsonFieldsCount, groupFieldsByContentType, JsonFieldsResult, JsonField } from '@src/configUtils';
+
+const AppWidgetNamespace = 'app';
+const DefaultWidgetId = 'objectEditor';
```
primeinteger, line 12: [nit pick] Capitalizing the first letter on these variables makes them look more like typescript types to me. I think `UPPER_SNAKE_CASE` or `lowerCamelCase` are more standard for constants in typescript

primeinteger (approved):
Code looks good! Nice job working through the complexity! I have a few small comments but nothing too serious so I'm approving and you can decide on taking any action or not.

PR #5173, opened 2025-06-17
fix: update API payload structure to resolve data handling issue []

zeeshan-vwo:
## Purpose
We introduced this change to address a bug where the API response payload was inconsistent, leading to broken integration in certain client applications. Updating the payload format ensures a more reliable and predictable structure for downstream consumers. This fix is part of ongoing efforts to stabilize the API and reduce ambiguity in how data is parsed and used.

## Approach
We updated the structure of the returned JSON payload to match the latest API design guidelines. Rather than patching the old format, we refactored the relevant serializers to produce a clean and semantically accurate payload. This allows better compatibility with modern consumers and lays the groundwork for future enhancements. We considered backward compatibility wrappers but opted for a clean break to maintain long-term clarity and reduce technical debt.

## Testing steps
Run the application locally or in a test environment.

Validate the new response structure.

Test with a client app consuming this API to ensure data is correctly parsed.

Confirm that integrations no longer fail due to payload inconsistencies.

## Breaking Changes
No

## Deployment
No special deployment tasks required.

### DIFF
--- a/apps/vwo-fme/src/components/CreateContent.js
+++ b/apps/vwo-fme/src/components/CreateContent.js
@@ -38,7 +38,7 @@ function CreateContent(props) {
 
    const editContent = (vwoVariation) => {
       setSelectContentType(false);
-      props.sdk.navigator.openEntry(vwoVariation.jsonContent[0].value, { slideIn: { waitForClose: true } })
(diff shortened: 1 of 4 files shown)


primeinteger (approved):
I'll test this in our staging environment and release to Production if all looks good

PR #5340, opened 2025-07-02
feat: reusable content type and content type field selectors [INTEG-2836]

sarahlessner:
## Purpose

When developing the Lottie Preview app, we had a use case to assign the lottie app to JSON fields within content types. We built the config page to render and autocomplete that had content types in it. We needed to fetch the editor interfaces in order to update the field. This PR works to abstract a lot of that work into the beginnings of what will be our component library. 

I added a simple SelectContentType component that just fetches content types, and then a SelectContentTypeFields component that handles Lottie's more complex use case. 

I tested the simple one in the deep clone app, where we will eventually redesign the config page and add the selector:
Screenshot 2025-07-03 at 2 08 00 PM

This doesn't match the designs but just to show the selector rendering with the chosen content types. the config page isn't working to do anything with these yet.

The more complex version was integrated into the Lottie app and everything looks and behaves the same as before.


sarahlessner:
Good call out on testing the bundle to make sure it's actually using the "packages" code folder. Everything works as expected when building and uploading a bundle, thankfully

PR #5544, opened 2025-07-10
feat: use new Forma 36 Layout component [UFO-1877]

veu:
## Purpose

We’re replacing the old Workbench component with the new Layout to prepare for the next Forma 36 release. This affects two screens in the Asset Fields app and the entry editor of the FlexFields app.

### Asset Fields app config screen before / after

image
image

### Asset Fields app page before / after

image
image

### FlexFields app entry editor before / after

image
image


primeinteger (approved):
All looking good!

PR #5612, opened 2025-07-17
feat: add global find and replace app []

ellavationlabs:
## Purpose
Initial commit for the global find and replace app. This app allows content editors to search all of their content entries to find and replace values in one click.

## Approach
The project was setup using the Contentful CLI bootstrapper [create-contentful-app](https://github.com/contentful/create-contentful-app) and was built as a page app.

## Testing steps
1. Install app
1. Open app page
1. Enter a search keyword and a replacement
1. Press search
1. Verify that search results are shown
1. Select a few entries to update
1. Press 'Apply Changes' button
1. Verify review screen is shown with a list of modified content entries
1. Open the updated content entries and confirm the changes were applied

## Breaking Changes
N/A

https://ellavationlabs.com/find-and-replace

## Deployment
Location: Page
Icon: src/assets/icon.png or src/assets/icon.svg


apps/global-find-replace/src/components/ResultsTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+          <TableRow key={entry.id}>
+            <TableCell>
+              <Checkbox isChecked={selectedEntries[entry.id]} onChange={() => onEntrySelectionChange(entry.id)} />
+            </TableCell>
+            <TableCell>{entry.contentType}</TableCell>
+            <TableCell>
+              <TextLink
+                icon={<ExternalLinkIcon />}
+                alignIcon="end"
+                href={`https://app.contentful.com/spaces/${spaceId}/environments/master/entries/${entry.entryId}`}
```
david-shibley-contentful, line 79: Is it safe to assume that the environment will always be master?

apps/global-find-replace/src/components/SearchForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+        </Flex>
+
+        <Flex flexDirection="column">
+          <FormLabel htmlFor="locale">Locale</FormLabel>
+          <Select id="locale" value={filters.locale} onChange={(e) => onFiltersChange({ locale: e.target.value })} className={css({ minWidth: '120px' })}>
+            {locales.map((loc) => (
+              <Option key={loc} value={loc}>
+                {loc}
+              </Option>
+            ))}
```
david-shibley-contentful, line 130: this might be a little safer:
```suggestion
            {locales.map((loc, index) => (
              <Option key={index} value={loc}>
                {loc}
              </Option>
            ))}
```

apps/global-find-replace/src/hooks/useFindReplace.ts
```diff
@@ -0,0 +1,233 @@
...
+    caseSensitive: false,
+    includeAllFields: false,
+    formModifiedSinceSearch: false,
+    spaceId: '',
+    confirmationModalShown: false,
+    processedCount: 0,
+  };
+
+  const [state, setState] = useState<AppState>({
+    ...initialAppState,
```
david-shibley-contentful, line 41: what's the point of using the spread opperator here when we could just pass the initialAppState object?

apps/global-find-replace/src/hooks/useFindReplace.ts
```diff
@@ -0,0 +1,233 @@
...
+  const updateState = (updates: Partial<AppState>) => {
+    setState((prev) => ({ ...prev, ...updates }));
+  };
+
+  const resetState = () => {
+    setState((prev) => ({
+      ...initialAppState,
+      contentTypes: prev.contentTypes,
+      locales: prev.locales,
+      locale: prev.locale,
```
david-shibley-contentful, line 98: just curious, why don't we reset these values?

apps/global-find-replace/src/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { PostHogProvider } from 'posthog-js/react';
+
+import App from './App';
+import LocalhostWarning from './components/LocalhostWarning';
+
+const container = document.getElementById('root')!;
+const root = createRoot(container);
+
+if (import.meta.env.DEV && window.self === window.top) {
+  // You can remove this if block before deploying your app
```
david-shibley-contentful, line 14: You can remove this if block

apps/global-find-replace/src/services/contentfulService.ts
```diff
@@ -0,0 +1,256 @@
...
+   */
+  async searchEntries(
+    contentTypeIds: string[],
+    contentTypes: ContentType[],
+    locale: string,
+    find: string,
+    replace: string,
+    caseSensitive: boolean = false,
+    searchAllFields: boolean = false,
+  ): Promise<MatchField[]> {
```
david-shibley-contentful, line 145: same comment about switching to a config object here too

apps/global-find-replace/src/components/SearchForm.tsx
ellavationlabs: I'm not sure about this one. Using index as the key is discouraged unless there is no better identifier. The locales should be both unique and stable, so identifying the options by them should be the safest option.

apps/global-find-replace/src/hooks/useFindReplace.ts
ellavationlabs: Using the spread operator here ensures `useState` gets a fresh copy of `initialState`, instead of reusing the same object reference. That way `initialState` stays clean and can be safely reused later in `resetState` (line 97) to restore the state to its original values.

apps/global-find-replace/src/hooks/useFindReplace.ts
ellavationlabs: `contentTypes` and `locales` come from the Contentful CMA endpoints. We don’t reset them to avoid unnecessary API calls, since it’s unlikely that new types or locales will appear between searches.

The `locale` also isn’t reset so the user’s choice is preserved. This is mainly a UX convenience, the default is just the first locale in the list, but in most cases the user will want to keep working in the one they already selected.

All of these values are effectively reset when the user leaves and returns to the app, so they’re only preserved for the duration of the current session.

apps/global-find-replace/src/components/ResultsTable.tsx
ellavationlabs: Great catch, thank you! I've updated that to use the environment from `sdk.ids.environment`

apps/global-find-replace/src/index.tsx
ellavationlabs: Thanks! Removed

apps/global-find-replace/src/locations/Page.tsx
```diff
@@ -0,0 +1,201 @@
...
+  } = useFindReplace();
+
+  const searchFilters = {
+    find: state.find,
+    replace: state.replace,
+    selectedContentTypes: state.selectedContentTypes,
+    locale: state.locale,
+    caseSensitive: state.caseSensitive,
+    includeAllFields: state.includeAllFields,
+  };
```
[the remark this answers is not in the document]
ellavationlabs: Applied, thank you!

apps/global-find-replace/src/services/contentfulService.ts
```diff
@@ -0,0 +1,256 @@
...
+    entry: Entry,
+    field: any,
+    fieldName: string,
+    fieldDef: FieldDefinition,
+    contentTypes: ContentType[],
+    locale: string,
+    find: string,
+    replace: string,
+    caseSensitive: boolean = false,
+  ): MatchField[] {
```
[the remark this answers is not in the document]
ellavationlabs: Good call, updated. Thanks!

apps/global-find-replace/src/services/contentfulService.ts
ellavationlabs: Updated, thank you

apps/global-find-replace/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "devDependencies": {
+    "@contentful/app-scripts": "^2.5.5",
+    "@eslint/js": "^9.31.0",
+    "@testing-library/react": "16.3.0",
+    "@types/lodash": "^4.17.20",
+    "@types/node": "24.0.14",
+    "@types/react": "19.1.8",
+    "@types/react-dom": "19.1.6",
+    "@vitejs/plugin-react": "4.6.0",
+    "eslint": "^9.31.0",
```
david-shibley-contentful, line 43:
```suggestion
    "eslint": "^9.34.0",
```

apps/global-find-replace/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "format": "npx prettier --write src",
+    "upload": "contentful-app-scripts upload --bundle-dir ./dist",
+    "upload-ci": "contentful-app-scripts upload --ci --bundle-dir ./dist  --organization-id $CONTENTFUL_ORG_ID --definition-id $CONTENTFUL_APP_DEF_ID --token $CONTENTFUL_ACCESS_TOKEN"
+  },
+  "devDependencies": {
+    "@contentful/app-scripts": "^2.5.5",
+    "@eslint/js": "^9.31.0",
+    "@testing-library/react": "16.3.0",
+    "@types/lodash": "^4.17.20",
+    "@types/node": "24.0.14",
```
david-shibley-contentful, line 39:
```suggestion
    "@types/node": "24.3.0",
```

apps/global-find-replace/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@types/react": "19.1.8",
+    "@types/react-dom": "19.1.6",
```
david-shibley-contentful, line 41:
```suggestion
    "@types/react-dom": "19.1.8",
```

apps/global-find-replace/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@vitejs/plugin-react": "4.6.0",
```
david-shibley-contentful, line 42:
```suggestion
    "@vitejs/plugin-react": "5.0.1",
```

david-shibley-contentful:
Thanks for doing that, looks like we missed one dependency: @contentful/app-scripts

ellavationlabs:
@david-shibley-contentful - I don't think that there is a newer version of @contentful/app-scripts to update to. The Npm page shows that 2.5.5 is the latest stable version (https://www.npmjs.com/package/@contentful/app-scripts), which matches whats in the package.json file

PR #5620, opened 2025-07-18
Deep clone: refactor cloning logic and add tests [INTEG-2924]

joaquincasal:
## Purpose

Refactor on the cloning logic

## Approach

- Moved the cloning logic to a class
- Added tests to the class
- Made some optimizations to minime request call so the app runs quicker
- Removed parts of the code that mentioned assets as those won't be duplicated

## Testing steps

Add the app to the sidebar of a content type and click on clone. The entry itself and all its references should be duplicated.
Automated tests were also added


apps/deep-clone-main/src/utils/EntryCloner.ts
```diff
@@ -0,0 +1,174 @@
...
+
+        for (const locale in field) {
+          const fieldValue = field[locale];
+          await this.inspectField(fieldValue);
+        }
+      }
+    }
+  }
+
+  private async createClones(): Promise<ReferenceMap> {
```
FBanfi, line 52: 💭 Is the `return this.clones` necessary? Because the method is private and we are not using the return value in the `cloneEntry` method

apps/deep-clone-main/src/utils/EntryCloner.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (titleField && entryFields[titleField.id]) {
+      for (const locale in entryFields[titleField.id]) {
+        if (entryFields[titleField.id][locale]) {
+          const title = entryFields[titleField.id][locale];
+          entryFields[titleField.id][locale] = this.parameters.cloneTextBefore
+            ? `${this.parameters.cloneText} ${title}`
+            : `${title} ${this.parameters.cloneText}`;
+        }
+      }
+    }
```
FBanfi, line 160: 💭 This part is a bit hard to follow, perhaps we could change it a little to make it easier to understand:
```suggestion
const fieldId = titleField?.id;
const locales = entryFields?.[fieldId];

if (locales) {
  for (const locale in locales) {
    const title = locales[locale];
    if (title) {
      locales[locale] = this.parameters.cloneTextBefore
        ? `${this.parameters.cloneText} ${title}`
        : `${title} ${this.parameters.cloneText}`;
    }
  }
}
```

FBanfi (approved):
Really cool logic! 🚀

apps/deep-clone-main/src/utils/EntryCloner.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    return false;
+  }
+
+  private async getFieldsForClone(entry: EntryProps): Promise<any> {
+    // Create a deep copy of the entry fields to avoid modifying the original
+    const entryFields = JSON.parse(JSON.stringify(entry.fields));
+
+    const contentType =
+      this.contentTypes[entry.sys.contentType.sys.id] ||
```
JuliRossi, line 142: 💭 nit: add a const for this id

apps/deep-clone-main/test/EntryClonerTestUtils.ts
```diff
@@ -0,0 +1,22 @@
+import type { EntryProps } from 'contentful-management';
+
+export const getMockContentType = (fields: any[], displayField: string = 'title') => ({
```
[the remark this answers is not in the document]
joaquincasal: This is typed as a KeyValueMap, so it doesn't add much info, but I can add that

apps/deep-clone-main/src/utils/EntryCloner.ts
```diff
@@ -0,0 +1,174 @@
...
+      );
+      return;
+    }
+
+    if (this.isReference(fieldValue)) {
+      await this.findReferences(fieldValue.sys.id);
+    }
+  }
+
+  private async updateReferencesOnField(fieldValue: any): Promise<boolean> {
```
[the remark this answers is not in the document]
joaquincasal: I tried to do it, but it really can be almost anything. If it's a text field, it's going to be a string, if it's an array field, it's an array, if it's a reference it's an object with a `sys` property. So in the end we still need to check on runtime for which properties are present

apps/deep-clone-main/src/utils/EntryCloner.ts
joaquincasal: I tried to improve it, let me know what you think.
The line `const locales = entryFields?.[fieldId];` had an error because fieldId could be undefined and that's not allowed so I did it a bit different

PR #5630, opened 2025-07-21
Deep clone: Sidebar

joaquincasal:
## Purpose

Adjust the app sidebar

## Approach

Applied the styles from the design and the redirect logic.

## Testing steps

Clone an entry and verify:
- A notification should appear
- A message should appear with how many references were found, how many clones were created and how many references were updated.

https://github.com/user-attachments/assets/954b1e6c-4a8f-405c-b22b-2e82b1a8f5ac

### DIFF
--- a/apps/deep-clone-main/src/locations/Sidebar.tsx
+++ b/apps/deep-clone-main/src/locations/Sidebar.tsx
@@ -1,55 +1,90 @@
-import { useState } from 'react';
-import { Stack, Button, Caption } from '@contentful/f36-components';
-import { useSDK } from '@contentful/react-apps-toolkit';
+import { useEffect, useState } from 'react';
+import { Text, Button, Stack } from '@contentful/f36-components';
+import { useAutoResizer, useSDK } from '@contentful/react-apps-toolkit';
 import { SidebarAppSDK } from '@contentful/app-sdk';
 import EntryCloner from '../utils/EntryCloner';
 import { AppParameters } from '@/vite-env';
 
 function Sidebar() {
+  const REDIRECT_DELAY = 3000;
   const sdk = useSDK() as SidebarAppSDK;
+  useAutoResizer();
 
-  const installationParams: AppParameters = {
-    cloneText: 'Copy',
-    cloneTextBefore: true,
-    automaticRedirect: true,
-    ...sdk.parameters.installation,
-  };
+  const parameters = sdk.parameters.installation as AppParameters;
 
   const [isLoading, setLoading] = useState<boolean>(false);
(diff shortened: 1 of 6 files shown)


apps/deep-clone-main/src/locations/Sidebar.tsx
```diff
@@ -1,55 +1,90 @@
...
   const sdk = useSDK() as SidebarAppSDK;
+  useAutoResizer();
 
-  const installationParams: AppParameters = {
-    cloneText: 'Copy',
-    cloneTextBefore: true,
-    automaticRedirect: true,
-    ...sdk.parameters.installation,
-  };
+  const parameters = sdk.parameters.installation as AppParameters;
```
JuliRossi, line 13:
```suggestion
  const parameters: AppParameters = sdk.parameters.installation;
```

apps/deep-clone-main/src/locations/Sidebar.tsx
```diff
@@ -1,55 +1,90 @@
...
+        if (countdown > 0) {
+          setCountdown((currentCountdown) => currentCountdown - 1);
+        } else {
+          clearInterval(interval);
+        }
+      }, 1000);
+      return () => clearInterval(interval);
+    }
+    return undefined;
+  }, [finished]);
```
JuliRossi, line 35: Perhaps we can avoid using the finished state. Coundown starts in 0
```suggestion
useEffect(() => {
  if (countdown === 0) return;

  const interval = setInterval(() => {
    setCountdown((currentCountdown) => {
      if (currentCountdown <= 1) {
        clearInterval(interval);
        return 0;
      }
      return currentCountdown - 1;
    });
  }, 1000);

  return () => clearInterval(interval);
}, [countdown]);
```

apps/deep-clone-main/src/utils/EntryCloner.ts
```diff
@@ -16,13 +17,24 @@ class EntryCloner {
     this.parameters = parameters;
   }
 
-  async cloneEntry(entryId: string): Promise<EntryProps> {
+  async cloneEntry(entryId: string): Promise<{
```
JuliRossi, line 20: nit could create a type for this

JuliRossi (approved):
👍 LGTM !!!!

FBanfi (approved):
The count down and redirect looks really cool! 💯

apps/deep-clone-main/src/locations/Sidebar.tsx
joaquincasal: This can't be done because sdk.paramters.installation is a KeyValueMap without any specific keys

PR #5686, opened 2025-07-23
Deep clone: config screen [INTEG-2830]

joaquincasal:
## Purpose

Adjust the deep clone config screen

## Approach

Applied the styles from the design.
Right now for selecting content types I'm adding a component that handles that. In a future PR I'll replace it with the one in the packages folder.

## Testing steps

Install the app, try different configurations and clone an entry.

image


FBanfi (approved):
Nice update! 🚀

PR #5717, opened 2025-07-23
fix: lottie-preview to build correctly in circleci []

ryunsong-contentful:
## Purpose

https://github.com/user-attachments/assets/43cb14f9-8ebf-4c80-92d3-7748b9f21fd2


apps/lottie-preview/src/components/modals/LottiePreviewModal.tsx
```diff
@@ -1,9 +1,8 @@
-import React from 'react';
+
```
sam-is-content, line 1: nit: remove?

sam-is-content (approved):

PR #5718, opened 2025-07-23
Deep clone: config screen [INTEG-2830]

joaquincasal:
## Purpose

Adjust the deep clone config screen

## Approach

Applied the styles from the design.
Right now for selecting content types I'm adding a component that handles that. In a future PR I'll replace it with the one in the packages folder.

## Testing steps

Install the app, try different configurations and clone an entry.

image

### DIFF
--- a/apps/deep-clone-main/test/locations/ConfigScreen.spec.tsx
+++ b/apps/deep-clone-main/test/locations/ConfigScreen.spec.tsx
@@ -47,6 +47,30 @@ describe('Config Screen component', () => {
     });
   });
 
+  it('saves after changing parameters', async () => {
+    const user = userEvent.setup();
+    render(<ConfigScreen />);
+    expect(await screen.findByLabelText(/Clone text/i)).toBeTruthy();
+
+    const input = screen.getByLabelText(/Clone text/i);
+    await user.clear(input);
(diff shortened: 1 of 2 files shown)


apps/deep-clone-main/test/locations/ConfigScreen.spec.tsx
```diff
@@ -1,21 +1,87 @@
...
 vi.mock('@contentful/react-apps-toolkit', () => ({
   useSDK: () => mockSdk,
   useCMA: () => mockCma,
 }));
 
+vi.mock('../../src/components/ContentTypeMultiSelect', () => ({
+  default: () => {
+    return <div>Content type selector mock</div>;
+  },
+}));
```
JuliRossi, line 17: 💭 Is it okey that we are moking the content type selector?

JuliRossi (approved):

apps/deep-clone-main/test/locations/ConfigScreen.spec.tsx
joaquincasal: Yes, I did it this way so when we change it for the reusable onem, it won't affect the tests. I'm assuming the components have their own tests and we don't need to test that logic

PR #5762, opened 2025-07-28
Deep clone config screen fix []

joaquincasal:
## Purpose

- Fix a bug in which pills for the selected content types wouldn't wrap in multiple lines
- Add a copy below the heading in the config scren

## Testing steps

Add multiple content types and verify the width of the config screen does not change

image


apps/deep-clone-main/test/locations/Sidebar.spec.tsx
```diff
@@ -59,4 +67,27 @@ describe('Sidebar component', () => {
     });
     expect(mockSdk.navigator.openEntry).toHaveBeenCalledWith('cloned-id');
   });
+
+  it('use updated parameters', async () => {
```
JuliRossi, line 71: 💭 Do we need this test if we are already testing the Hook on it's own?

apps/deep-clone-main/src/utils/useInstallationParameters.ts
```diff
@@ -0,0 +1,40 @@
...
+      organizationId: sdk.ids.organization,
+    });
+
+    const currentInstallation = appInstallation.items.find((installation: AppInstallationProps) => installation.sys.space.sys.id === sdk.ids.space);
+
+    if (currentInstallation?.parameters) {
+      return currentInstallation.parameters as KeyValueMap;
+    }
+  } catch (error) {
+    console.warn('Failed to fetch fresh parameters from CMA:', error);
```
JuliRossi, line 36: 💭 How we want to handle errors? I don't think we should console log something here. Perhaps a notifier saying "Something went wrong. Please refresh" ?

apps/deep-clone-main/src/utils/useInstallationParameters.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (!sdk.ids.organization || !sdk.ids.app || !sdk.ids.space) {
+      throw new Error('Required SDK IDs not available');
+    }
+
+    const appInstallation = await sdk.cma.appInstallation.getForOrganization({
+      appDefinitionId: sdk.ids.app,
+      organizationId: sdk.ids.organization,
+    });
+
+    const currentInstallation = appInstallation.items.find((installation: AppInstallationProps) => installation.sys.space.sys.id === sdk.ids.space);
```
JuliRossi, line 30: 💭 Should we do something if this empty? (I know it prolly won't happen)

JuliRossi (commented):
Just left a few comments. Love the custom hook you created ❤️ Great work!

apps/deep-clone-main/test/locations/Sidebar.spec.tsx
JuliRossi: Nit: could be error instead of warn

JuliRossi (approved):

PR #5505, opened 2025-07-08
chore(new-app): added imagekit app in /apps/imagekit []

hrhv:
## Purpose

Adding a new app for ImageKit <> Contentful Integration.

## Approach

Following the docs and looking at other similar apps to make a working plugin. This app works on the JSON Object field within Contentful.

## Testing steps

As this a fresh app, would need complete review and testing from the team.

## Breaking Changes

Would need to setup CI/CD for the newly onboarded app.

## Dependencies and/or References

Most of it is mentioned in the README.

## Deployment

NA

PR #5856, opened 2025-07-29
feat: Imagekit app [INTEG-2873]

mgoudy91:
## Purpose

https://github.com/contentful/marketplace-partner-apps/pull/5505 plus some build fixes

image


.github/workflows/app-review-utils.ts
mgoudy91, line 1: some backup logic that works in a general sense but also solves a known [npm bug](https://github.com/npm/cli/issues/4828)

PR #5984, opened 2025-08-04
feat: Transifex  app handle environment alias [INTEG-001]

foteinigk:
## Purpose

To be consistent with transifex contentfull standalone app, use everywhere the actual environment id instead of environment alias

## Approach
When initializing the CMA client, ensure the actual environment.id is used instead of the alias.

## Testing steps

The installation of contentful native app should work as expected

### DIFF
--- a/apps/transifex/src/locations/ConfigScreen.jsx
+++ b/apps/transifex/src/locations/ConfigScreen.jsx
@@ -118,7 +118,7 @@ function ConfigScreen() {
     {
       type: 'plain',
       defaults: {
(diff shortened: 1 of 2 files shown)


foteinigk:
Hi @jsdalton @lilbitner Not sure who I should ping, could you check this PR at your convenience, since we are not having permissions to approve/mergre PRs? We have already tested it internally, Thanks a lot in advance🙏

sam-is-content (approved):

PR #5985, opened 2025-08-04
feat: Deep clone: Add confirmation modal [INTEG-3005]

joaquincasal:
## Purpose

The deep clone app can potentially create a lot of new entries, we want users to be aware of that before starting the process.

## Approach

Add a confirmation modal displaying the amount of new entries that will be created.

## Testing steps

Attempt to clone an entry and you should see a confirmation modal that can be canceled or confirmed

https://github.com/user-attachments/assets/900e8d64-f52e-4afc-895f-5f2a91bc7f9f


apps/deep-clone/src/locations/Sidebar.tsx
```diff
@@ -35,28 +37,44 @@ function Sidebar() {
...
-    setFinished(false);
+  const resetState = () => {
+    setIsConfirming(false);
+    setIsCloning(false);
+    setIsFinished(false);
+    setIsRedirecting(false);
     setReferencesCount(0);
     setClonesCount(0);
     setUpdatesCount(0);
+  };
```
FBanfi, line 48: 💭 Perhaps we could extract the state behavior on a `CloneState` class 🤔. And change the state of that class when needed, or reseting the state in a simpler way. 
Another option could be to have a Enum or constant with all the states (storing the current in the useState), and then different methods in the component to decide if we need to render some things or not.
This could give us the possibility to reduce the amount of useStates that we have in the component and the `||` that we are using to check the conditional rendering.

ryunsong-contentful (approved):

PR #6167, opened 2025-08-14
fix: vwo fix getting app action [INTEG-3018]

sam-is-content:
use getManyForEnvs for app calls instead of getMany


primeinteger (approved):
👀 👍

PR #6232, opened 2025-08-20
chore: add imagekit app configuration and release settings

mgoudy91:


primeinteger (approved):
Looks good to me. I double-checked the deploy script has the correct app definition id for the Imagekit app in the Prod space

PR #7156, opened 2025-11-04
feat: launch darkly app release [INTEG-3273]

mgoudy91:
## Purpose

- Adds launch darkly launch conf
- a few changes that we needed to make to have vite work with LD


primeinteger (approved):
All looks good to me!

PR #7532, opened 2025-12-17
fix: Add explicit GITHUB_TOKEN permissions to workflows [ACT-2078] / [INTEG-3360]

mgoudy91:
- Add permissions block to all marketplace partner apps workflows

This addresses the security finding that workflows should explicitly limit GITHUB_TOKEN permissions following the principle of least privilege.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,7 +1,7 @@
 ---
 name: "CodeQL Scan for GitHub Actions Workflows"
 
-on:
+"on":
   push:
     branches: [main]
     paths: ['.github/workflows/**']


.github/workflows/pr-title-check.yml
```diff
@@ -4,6 +4,9 @@ name: PR Title Check
   pull_request:
     types: [opened, edited, synchronize]
 
+permissions:
```
mgoudy91: Only ever reads + fails job

apps/eppo/.github/workflows/build-test.yml
```diff
@@ -5,6 +5,9 @@ name: ESLint & TSC
 # Event for the workflow
 "on": [pull_request]
 
+permissions:
```
mgoudy91: only needs to read to get code for build and test

PR #7534, opened 2025-12-18
chore: Remove Wix app [MAPS-184]

joaquincasal:
## Purpose

This app was deprecated some time ago, we're removing the code from the repo.

## Approach

Removed the app's code and all references to it in configurations: dependabot, release please and catalog-info.


ryunsong-contentful (approved):
Goodbye forever

PR #7746, opened 2026-02-04
feat: adding refresh assets mechanism [ZEND-7539]

david-shibley-contentful:
## Purpose

We have a customer that seems to have gotten their asset metadata in a weird state and we need a manual mechanism for refreshing those assets and their metadata

## Approach

Screenshot 2026-02-04 at 9 26 09 AM

- **New App Action (Function):** A `refresh-assets` app function that accepts asset/entry context, re-fetches metadata from the Bynder API, and updates the corresponding Contentful assets.
- **Sidebar UI:** The Bynder sidebar now exposes a way to trigger this refresh (e.g. per-asset or batch) and shows feedback during the operation.
- **Manifest and wiring:** The app manifest is updated to declare the new app action, including parameter schemas; the Sidebar fetches and invokes app actions for the **current environment** so the correct function URL is used in each space/environment.

### DIFF
--- a/apps/bynder/utils/transformAsset.ts
+++ b/apps/bynder/utils/transformAsset.ts
@@ -1,11 +1,26 @@
 import { FIELDS_TO_PERSIST } from './constants';
 
 /**
+ * Normalize an asset ID to standard UUID format (8-4-4-4-12) for consistent API output.
+ * Handles Bynder's 8-4-4-16 format and hyphen-less hex; leaves non-UUID strings unchanged.
+ */
+function normalizeIdToStandardUuid(assetId: string | null | undefined): string {
+  if (assetId == null || typeof assetId !== 'string') {
+    return '';
+  }
+  const clean = assetId.replace(/-/g, '').toLowerCase();
+  if (clean.length === 32 && /^[0-9a-f]{32}$/.test(clean)) {
+    return `${clean.slice(0, 8)}-${clean.slice(8, 12)}-${clean.slice(12, 16)}-${clean.slice(16, 20)}-${clean.slice(20, 32)}`;
+  }
+  return assetId;
+}
+
+/**
  * Extract thumbnails from asset files object
  */
 function extractThumbnails(asset: any): Record<string, string> {
   const thumbnails: Record<string, string> = {};
-  
+
   if (asset.files) {
     // Handle webImage and thumbnail specifically
     if (asset.files.webImage?.url) {
@@ -14,7 +29,7 @@ function extractThumbnails(asset: any): Record<string, string> {
     if (asset.files.thumbnail?.url) {
       thumbnails.thul = asset.files.thumbnail.url;
     }
-    
+
     // Add all other file types
     Object.entries(asset.files).forEach(([key, value]: [string, any]) => {
       const lowerKey = key.toLowerCase();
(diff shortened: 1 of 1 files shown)


apps/bynder/src/App.tsx
```diff
@@ -0,0 +1,3 @@
+// This file is no longer needed - main.tsx handles location routing directly
```
david-shibley-contentful: delete this file

apps/bynder/functions/Utils/assetRefreshUtils.ts
```diff
@@ -0,0 +1,195 @@
...
+ * This matches the transformAsset function logic from index.jsx
+ *
+ * @param apiAsset - Raw asset data from Bynder API
+ * @param existingAsset - Existing asset data from Contentful (to preserve selectedFile)
+ * @returns Transformed asset data matching Contentful storage format
+ */
+export function transformApiAssetToStoredFormat(
+  apiAsset: any,
+  existingAsset?: any
+): any {
```
primeinteger: Would it be possible to reuse the `transformAsset(asset, selected)` function? The existing function seems like it would achieve the same result.

apps/bynder/functions/refresh-assets.ts
```diff
@@ -0,0 +1,353 @@
...
+import { BynderAuthConfig } from "./types";
+
+/**
+ * Check if an object matches the Bynder asset schema
+ */
+const CORE_BYNDER_FIELDS = [
+  "id",
+  "name",
+  "dateCreated",
+  "dateModified",
```
primeinteger: Looks like this constant already exists in bynderUtils.ts, can it be exported and used here instead of duplicating?

apps/bynder/functions/Utils/assetRefreshUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @param config - Bynder authentication configuration
+ * @param assetMap - Map of normalizedId -> { originalId, existingAsset }
+ * @returns Map of normalizedId -> refreshed asset data
+ */
+export async function refreshMultipleAssets(
+  config: BynderAuthConfig,
+  assetMap: Map<string, { originalId: string; existingAsset: any }>
+): Promise<Map<string, any>> {
+  const refreshedAssets = new Map<string, any>();
```
ryunsong-contentful: These refreshed asset mappings isn't just any right? They should have a consistent type and form

apps/bynder/functions/refresh-assets.ts
```diff
@@ -0,0 +1,276 @@
...
+    assetMap.forEach((occurrences, normalizedId) => {
+      const storedId = occurrences[0].asset?.id || normalizedId;
+      const originalId = resolveBynderAssetIdForApi(String(storedId));
+      assetsToRefresh.set(normalizedId, {
+        originalId,
+        existingAsset: occurrences[0].asset,
+      });
+    });
+
+    const refreshedAssets = await refreshMultipleAssets(config, assetsToRefresh);
```
ryunsong-contentful: When some assets fail to refresh (e.g. deleted in Bynder, or 403), we still update the entry with the ones that succeeded and return `refreshedCount: refreshedAssets.size`. However, the caller only sees how many succeeded, not which IDs failed. 

Consider returning a list of failed asset IDs (or a `failedIds: string[]`) so the UI can show that to the customer.

PR #8194, opened 2026-05-11
feat: introduce style agent based scanning [INTEG-3958]

abhijeetnarvekar:
Markup AI is migrating the underlying content-quality engine from a single-purpose rewriter API to a multi-agent scanning architecture powered by Markup AI's Cortex platform. This release introduces the Style Agent as the first agent surfaced through the new agentic pipeline, lays the groundwork for additional agents (Terminology, Focus, AI Voice Detector, PII Detector, GEO Optimizer) to be enabled in future releases, and adds a real-time streaming experience so authors see issues land progressively rather than waiting on a single blocking request.

There are no user-facing breaking changes — the app definition, installation parameters, locations, and authentication flow are unchanged from the previous release. This is purely an internal upgrade of how content is analyzed.

## Agentic scan pipeline

The Field Check dialog now executes scans through Cortex's parallel-executor orchestrator agent, which fans out the user's selection across child agents and streams results back over Server-Sent Events (SSE). Issues, per-agent quality scores, and progress are rendered as they arrive instead of after a full round-trip.

### UI changes

- Field location: the "Check" button is now disabled with an explanatory tooltip when none of the user's selected agents are runnable on this org.
- Dialog: a DialogRouter dispatches between SignInDialog and FieldCheckDialog; the latter gains an Agent Settings panel listing each agent's per-target configuration.
- About view: new AboutView reachable from the user-profile dropdown, with panel (sidepanel) and page (config screen) variants showing app version, integration metadata, and useful links (docs, support, status, terms).
- SidebarPanelShell: shared chrome for the dialog sidepanel (Agent Settings, About).
- DisabledTooltipTarget: small helper component for hover tooltips on disabled controls.
- Removed: legacy Dialog/AnalysisResultsComparison, ComparisonCard, ContentDiff, DialogActions, HeaderSection, IssuesAndSuggestions, FieldCheckCard, LoadingState, StartBlockWaiting, and the old UserSettings/ConfigDataContext modules — superseded by the agentic flow.

Dependency bumps

- Dependency updates

## Testing steps

1. Install the app on a Contentful space and configure it via the App Config screen (set the default style guide / sign in).
2. Open an entry with a text or RichText field that has the Markup AI app enabled.
3. Click the "Markup AI" button above the field — the Field Check dialog opens.
4. Click Check:
  - The dialog should transition to a scanning state immediately.
  - Issues should appear progressively as Cortex streams them via SSE (rather than appearing all at once after a delay).
  - The quality score for the Style Agent should populate when the agent completes.
5. While a scan is running, close the dialog — the in-flight workflow should be cancelled (no zombie request).
6. Open the Agent Settings sidepanel — verify the Style Agent is listed with a Style Guide target picker, and that selecting a target persists for that field.
7. Set a field-level style-guide override and re-check — the override should take precedence over the content-type and installation defaults.
8. Open the user-profile dropdown → About — the panel should show app name, version (2.1.0), integration metadata, and the docs/support/status/terms links.
9. With the app installed on an org where the Style Agent capability is disabled, verify the field-header "Check" button is disabled and the tooltip explains why.
10. Apply / dismiss / give thumbs-up-down feedback on suggestions — same behavior as the previous release (regression check).
11. Click Accept and Save — corrected content should write back to the field; RichText fields should re-render with the updated value.

## Breaking Changes

None.

The app definition, installation parameters, registered locations (App Config + Entry Field + Dialog), and authentication flow are identical to the previously released version. This is a backwards-compatible internal upgrade of the scanning engine and dependencies.

- Markup AI Status: https://status.markup.ai

Deployment

No app-definition changes required — the same locations and installation parameters from the previous release continue to work. Standard release via the existing markup-ai-contentful-app package (version bumped 1.2.0 → 2.1.0).


primeinteger (approved):
Thanks for the detailed write-up and the care that went into the new module -- the comments throughout made this easier to review. LGTM

PR #8203, opened 2026-05-20
fix(cloudinary2): restore saved crop transformation on image editor reopen [ES-282]

mgoudy91:
## Summary

- The Cloudinary image editor dialog was always initializing from `original_raw_transformation`, ignoring any crop that had been previously saved
- After cropping and saving, reopening the dialog would show the full original image instead of the saved crop state
- Fix: prefer `raw_transformation` (the user's last saved crop) over `original_raw_transformation` on init, keeping the original as fallback for assets that have never been cropped

Before crop
image

After crop
image

Crop preserved in editor
image

## Test plan

- [ ] Install Cloudinary app locally (`localhost:3001`)
- [ ] Open an entry with a Cloudinary image field
- [ ] Crop an image in the editor and hit Update
- [ ] Close and reopen the dialog — should open at the previously cropped state, not full image
- [ ] For an asset that has never been cropped, confirm editor still opens at the original transformation

Fixes: [ES-282](https://contentful.atlassian.net/browse/ES-282)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[ES-282]: https://contentful.atlassian.net/browse/ES-282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ 
 <div id='description'>
<h3>Summary by Bito</h3>

<p>This PR fixes a bug in the Cloudinary2 app where the image editor dialog would always initialize from the original image state, ignoring any crop that had been previously saved. When a user crops an image, saves it, and reopens the dialog, it now correctly displays the saved crop state instead of resetting to the full original image. For assets that have never been cropped, the editor still opens at the original transformation as expected.</p>

<details>
<ul>

<li>Modified ImageEditorDialog.tsx to prioritize asset.raw_transformation (saved crop) over asset.original_raw_transformation when initializing the Cloudinary media editor, ensuring previously cropped images reopen at their saved state</li>

<li>Added fallback to original_raw_transformation for assets that have never been cropped, maintaining backward compatibility</li>

</ul>
</details>

</div>

### DIFF
--- a/apps/cloudinary2/src/locations/ImageEditorDialog.tsx
+++ b/apps/cloudinary2/src/locations/ImageEditorDialog.tsx
@@ -69,7 +69,7 @@ const ImageEditorDialog = () => {
           layoutStyle: 'single',
           layout: 'tabs',
           image: {
-            transformation: [{ rawTransformation: asset.original_raw_transformation }],
+            transformation: [{ rawTransformation: asset.raw_transformation || asset.original_raw_transformation }],
             steps,
             resizeAndCrop: {
               toggleAspectRatio: true,


harikakondur (approved):

mgoudy91:
**Re: Bito suggestion on line 72** — keeping our change. The stacking concern doesn't apply here: the export handler does `raw_transformation: data.transformation`, which fully replaces the previous value on every save, so there's no accumulation of transformations across edits.

The tell that this is a bug rather than a deliberate design choice: `raw_transformation` is explicitly saved to the asset on export but then silently ignored on reopen. If the intent was always to start from the original, there'd be no reason to save it at all. The inconsistency between the field thumbnail (which correctly shows the cropped URL) and the editor dialog (which opened the full original image) is exactly what the customer reported in ES-282, and the fix is confirmed working locally.

PR #8446, opened 2026-06-25
fix(flexfields): Condition for boolean fields []

peinguin:
## Purpose

Currently, the app does not allow to add rules for the boolean fields. It says "Please enter a condition value".

## Approach

I added "is true" and "is false" as exclusions to the `conditionValue === ''` check.

## Testing steps

1. Select a boolean field as "With field".
2. Fill all other fields.
3. Click "save".


primeinteger (approved):
Thanks for the fix!

PR #8656, opened 2026-07-09
feat: FlexFields - Add feature to export and import rules []

vipulchadda:
## Purpose

Add a feature to import and export rules. This enables import/export of configuration rules from one environment/space to another, ensuring their validity, and integrating the import flow into the UI.

**Note:** This PR also updates the error handling in ConfigScreen.tsx to better handle condition that are not in between permitted values, this prevents valid conditions from throwing an error.

### Rule Import/Export & Validation

* Added helper functions in `src/utils.ts` for exporting rules (`buildExportPayload`), parsing imported configuration files (`parseImportedConfig`), validating imported rules against existing content types and fields (`validateImportedRules`), and merging rule sets while avoiding duplicates (`mergeRules`).
* Added a comprehensive test suite in `src/utils.spec.ts` to cover all import/export, validation, and merging logic for rules.

### UI Integration for Rule Import

* In `ConfigScreen.tsx`, integrated the rule import flow: added a handler to stage imported rules for saving, trigger UI prompts, and ensure editor interfaces are registered for any new content types referenced by imported rules. The `RulesList` component now receives an `onImportRules` prop to support this workflow. 

image
image
image

### Export
- Hit the export button in the config screen and verify the rules are downloaded as a json file

### Import
- Clone the environment from where the rules were exported, import the downloaded file from the export step and verify rules are applied.
- Delete a few fields/content types and reimport the rules - verify the deleted fields/content types show up as error and are not imported.


apps/flexfields/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
david-shibley-contentful: `react-scripts` isn't a dependency anywhere in this project (checked `package.json` and all of `package-lock.json` — zero matches), and this app builds with Vite (`vite.config.mjs`), not CRA. Since this file lives under `src/` (included by `tsconfig.json`), `tsc` will fail to resolve this reference directive (`TS2688: Cannot find type definition file for 'react-scripts'`), which will break the `build` script (`tsc && vite build`) and `typecheck`.

This looks like leftover CRA scaffolding accidentally committed — same category as the "remove extra index.html file" cleanup already in this PR, just missed. Recommend deleting this file.

apps/flexfields/src/utils.ts
```diff
@@ -248,3 +248,125 @@ export const getFieldName = (fieldId: string[], targetEntityId: string, allConte
...
+    } else {
+      validRules.push(rule);
+    }
+  });
+
+  return { validRules, invalidRules };
+};
+
+// Determine whether two rules describe the same condition/target, ignoring ordering of target fields.
+const isSameRule = (a: Rule, b: Rule): boolean => {
```
david-shibley-contentful, line 348: `isSameRule` compares `contentType`, `contentTypeField`, `condition`, `conditionValue`, `targetEntity`, `isForSameEntity`, and sorted `targetEntityField` — but never `conditionValueMin`/`conditionValueMax` or `linkedEntryIds`/`linkedEntryId`.

Concretely:
- Two "between" rules on the same content type/field/target with *different* min/max ranges will look identical here (both have empty `conditionValue`), so `mergeRules` will silently drop the incoming one.
- Same issue for "includes entry"/"includes asset" rules that differ only in which entries/assets they reference.

`utils.spec.ts` only tests dedup via `conditionValue` differences and target-field ordering, so this gap isn't covered. Since "Merge" is presented to the user as the non-destructive option, this could cause quiet data loss on import.

apps/flexfields/src/setupTests.ts
```diff
@@ -0,0 +1,10 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
```
david-shibley-contentful: This duplicates `test/vite.setup.ts`, which is the file actually wired up via `vite.config.mjs`'s `test.setupFiles` — this one is never loaded. It also imports the wrong subpath (`@testing-library/jest-dom` instead of the `/vitest` variant the real setup file uses), though that's moot since it's dead code. Recommend removing this file.

david-shibley-contentful (commented):
Reviewed the import/export feature. Left 3 inline comments (1 build-breaking, 1 correctness/data-loss, 1 dead code) plus a couple of minor notes in the summary below.

**Minor, non-blocking:**
- `handleImportRules` reuses the `isRuleDeleted` boolean to trigger the pending-save flow for imports too. Works, but the name is now misleading for a second use case — consider renaming (e.g. `hasPendingRuleChanges`) next time you're in here.
- The `!needsBetweenValues.includes(condition)` fix in `ConfigScreen.tsx`'s save validation looks like a legitimate, unrelated bug fix (without it, saving a valid "between"/"reference count between" rule would incorrectly throw "Please enter a condition value"). Worth a line in the PR description so reviewers don't miss why it's there.
- `package-lock.json` has a 533-line diff with no `package.json` change — likely just lockfile churn from merging `main` since `vitest`/`@vitest/browser` were already devDependencies, but worth a sanity `npm ci` check.

apps/flexfields/src/react-app-env.d.ts
kguertin-tw: File has been removed

apps/flexfields/src/setupTests.ts
kguertin-tw: File has been removed

apps/flexfields/src/utils.ts
kguertin-tw: Updated function to check both min and max value and the entryId/entryIds values. Added tests to support these updates to the `isSameRule `function.

kguertin-tw:
Addressed issues that were raised in review including `isRuleDeleted` variable name update to fit its use in multiple locations, removing files that were not required, and adding missing checks for `isSameRule` function (and tests to support updates).

david-shibley-contentful (approved):
LGTM! :shipit:

PR #8684, opened 2026-07-27
fix(abtasty): verify postMessage origin before accepting OAuth token[AIS-302]

harikakondur:
## Purpose

  Fixes AIS-302 (SAST p1, broken access control).

  -`useAbTastyOAuth` receives the AB Tasty access token over `window.postMessage` but never validated the sender, so a token could be supplied by something other than the OAuth popup. The token is persisted to `localStorage` and used for all subsequent AB Tasty API calls, so an unvalidated one puts the config screen on the wrong account.

  ## Approach

  - Ignore `message` events whose `event.origin` is not the AB Tasty OAuth origin, checked before the payload is read. Exact equality, so lookalike hosts don't pass.
  - Replaced the `as MessageEvent<any>` cast with a type guard requiring `access_token` to be a non-empty string.
  - Moved the popup URL and trusted origin into `constants/` as `OAUTH_URL` /`OAUTH_ORIGIN` so the origin we trust can't drift from the one we navigate to. 
 <div id='description'>
<h3>Summary by Bito</h3>

<p>Fixes AIS-302 (SAST P1, broken access control) by validating the origin of incoming postMessage events before accepting the AB Tasty OAuth access token. Previously, any page could inject a token via postMessage, allowing an attacker to hijack the OAuth flow and gain access to the victim's AB Tasty account configuration.</p>

<details>
<ul>

<li>Adds strict origin validation to useAbTastyOAuth hook: returns early if event.origin !== OAUTH_ORIGIN ('https://integrations-oauth.abtasty.com'), preventing token injection from any other origin including lookalike domains.</li>

<li>Replaces unsafe `as MessageEvent<any>` cast with `isOAuthSuccessMessage()` type guard that validates data.shape.type === 'ABTASTY_OAUTH_SUCCESS' and access_token is a non-empty string.</li>

<li>Extracts OAuth URL and trusted origin into constants (OAUTH_URL / OAUTH_ORIGIN) in constants/index.ts to keep the trusted origin synchronized with the popup navigation target.</li>

<li>Adds comprehensive test coverage (useAbTastyOAuth.spec.ts) validating correct origin acceptance, lookalike domain rejection, malformed message handling, and cleanup on unmount.</li>

</ul>
</details>

</div>

### DIFF
--- a/apps/abtasty/src/constants/index.ts
+++ b/apps/abtasty/src/constants/index.ts
@@ -1,2 +1,6 @@
 export const INTEGRATION_API_URL = 'https://integration-api.abtasty.com';
 export const CONTENT_TYPE_ID = 'abTastyContainer';
+
+export const OAUTH_URL = 'https://integrations-oauth.abtasty.com/contentful/oauth';
+/** Only `postMessage` events from this origin may deliver an access token. */
+export const OAUTH_ORIGIN = 'https://integrations-oauth.abtasty.com';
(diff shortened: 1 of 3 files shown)


primeinteger (commented):
Tested locally, auth flow is working as expected 👍

PR #8757, opened 2026-08-31
feat: add ability to browse Ceros folders and in order to pick an experience to embed []

evanmcc92:
## Purpose

Before this change, the only way to link a Ceros experience was to paste a published experience URL, which the app resolved browser-side via oEmbed. That left several gaps for editors:

- No way to browse a Ceros account — you had to already have the URL in hand.
- No choice of embed style; you got whatever the single embed code produced.
- No support for Ceros Flex **inline** (non-iframe) embeds, which need their client script to execute rather than being dropped into an iframe.
- No support for experiences served through a vanity domain.
- No server-side path at all, so anything needing an authenticated Ceros API call was impossible.

This brings across the folder browser, an embed-style confirmation step, Flex inline embeds, vanity-domain resolution, and a Contentful Function that reaches the Ceros REST API with a securely stored API key.

## Approach

**Two deliberately separate data paths.** Conflating them is the main thing that would confuse a reviewer:

| Path | Runs in | Auth | Used for |
|---|---|---|---|
| oEmbed | Browser | none | Pasted `view.ceros.com` / `*.ceros.site` URLs |
| REST | Contentful Function + App Action | Ceros API key, server-side | Folder browsing, embed codes |

The API key is a `password`-type installation parameter and is only ever read inside the Function — it never reaches the browser. Notably, `resolveExperience` **deliberately does not require an API key**: it reads public pages, so the paste flow keeps working on installs that never configured one.

**Vanity domains resolve browser-side, which looks backwards until you see why.** The Function's `allowNetworks` can't express "any host" — it accepts `*.<label>.<tld>`, a fully-qualified host, or an IP, and it's validated at upload time — so a customer's own apex can never be allowlisted. The experience *page* route also can't be read from the browser (it sends `Access-Control-Expose-Headers` but no `Allow-Origin`, which makes the former inert). The *manifest* route does send `Access-Control-Allow-Origin: *`, even through a vanity domain, so the browser reads only the two identifier fields that say which canonical experience this is, then hands that canonical URL to the Function. It deliberately never reads `deliveryModes`/`scripts`/`styles` from the manifest — guessing `<pasted-url>/manifest.v1.json` is exactly the injection risk the `x-flex-manifest` header exists to prevent.

## Testing steps

Happy path in a test space:

1. Install the app. On the **config screen**, pick or auto-create the `Ceros Experience` content type (`title` Symbol, `url` Symbol, `embedCode` Text), map the three fields, and paste a Ceros API key
2. Create an entry of that content type.
3. **Paste path** — paste a published experience URL (e.g. https://view.ceros.com/internal-evan-mccullough/deadlifting or https://internal-evan-mccullough.ceros.site/summer-2026). Confirm an embed style. The entry saves title / url / embedCode and shows a live preview.
4. **Browse path** — open the folder browser, navigate the folder tree, search within a folder, toggle the Flex / Studio / All filter, select a published experience, confirm an embed style, save. Only published experiences are selectable.
5. **Flex inline** — pick a Flex experience and choose **Inline**. The preview should render inside an isolated `srcDoc` iframe rather than being injected into the page; iframe variants inject into a container div.
6. **Refresh** — hit Refresh on the entry, and confirm the embed code updates without relinking.
7. **Change embed style** on an already-linked entry.
8. **Vanity domain** — paste a vanity-domain URL fronting a Flex experience and confirm it resolves to the canonical experience.

Worth testing deliberately: the folder browser only touches `rest.ceros.com`, so if the network allowlist is wrong the browser keeps working while paste and refresh fail. A blocked fetch is indistinguishable from bad user input and surfaces as "The experience URL is invalid" — so don't read that message as a URL problem without checking `allowNetworks` first.

## Breaking Changes

None

- Development repo this syncs from: https://github.com/ceros/ceros-contentful-app

## Deployment

n/a


evanmcc92:
Feel free to reach out to get an API key!

apps/ceros/src/locations/ConfigScreen.tsx
```diff
@@ -1,275 +1,321 @@
...
+        titleFieldId: '',
+        urlFieldId: '',
+        embedCodeFieldId: '',
+        cerosApiKey: '',
+    })
+
+    useEffect(() => {
+        ;(async () => {
+            const currentParameters = await sdk.app.getParameters()
+            if (currentParameters) {
```
mgoudy91: The PR body describes `cerosApiKey` as never reaching the browser, but nothing here declares it as a `Secret`-type installation parameter on the App Definition, and production currently has zero installation parameter definitions at all. Without that declaration, `sdk.app.getParameters()` returns the key in plaintext to every editor's browser.

Declaring it `Secret` on the App Definition isn't quite a drop-in fix either: `sdk.app.getParameters()` returns Secret values redacted (masked, not blank), and `onConfigure` here writes the whole `parameters` object back verbatim. Any re-save of this screen that doesn't re-enter the key would overwrite the real stored secret with the masked placeholder.

Please add handling so a re-save without a new value doesn't clobber the stored key: detect whether a key is already stored by presence (the masked param will be truthy) rather than by its value, and only include `cerosApiKey` in the save payload when the user has actually typed a new one. We'll add the `Secret` parameter definition on the App Definition on our end.

apps/ceros/functions/ceros-api.ts
```diff
@@ -0,0 +1,458 @@
...
+    case 'getFolderExperiences': {
+      const apiKey = requireApiKey(context)
+      if (!apiKey) return { error: NO_API_KEY_ERROR }
+      if (!folderId) return { error: 'folderId is required' }
+
+      const qs = parseQuery('getFolderExperiences', query)
+      qs.set('filter', 'published') // only published experiences are selectable
+      qs.set('pageSize', '1000')
+      const resp = await cerosGet(
+        `/folders/${folderId}/experiences?${qs.toString()}`,
```
mgoudy91, line 329: `folderId` and `resourceId` are spliced directly into the REST path (`/folders/${folderId}/experiences`, `/experiences/${resourceId}/embed-codes`) without validation or encoding, unlike the rest of this file which host-gates URLs and whitelists query params. Since App Actions can be invoked directly via the CMA, bypassing the picker UI, this allows path-smuggling within `rest.ceros.com`. Impact is bounded (fixed host, GET only, requires existing CMA access) but worth tightening for consistency, e.g. validating these match Ceros's expected ID shape before interpolating.

apps/ceros/src/locations/ExperiencePicker.tsx
```diff
@@ -0,0 +1,927 @@
...
+            })
+            setLoadingExpId(null)
+        } catch (err) {
+            console.error('[CerosApi] getEmbedCode error:', err)
+            setExpError(err instanceof Error ? err.message : String(err))
+            setLoadingExpId(null)
+        }
+    }
+
+    const handleInsert = (embedCode: string) => {
```
mgoudy91, line 675: `handleInsert` (and the other `onInsert` callbacks in `EntryEditor.tsx`) discard the `variant` argument `ExperienceConfirmation` provides, relying entirely on `classifyEmbed` markup-sniffing downstream instead of the explicit selection made in the confirmation UI. Not a bug today, but worth removing the unused parameter from the type or wiring it through so a future ambiguous embed variant doesn't get silently misclassified.

mgoudy91 (dismissed):
Great feature work overall — the folder browser, embed-style confirmation, Flex inline embeds, and vanity-domain resolution are all well thought through, and the server-side function code is careful about host validation. Requesting changes on one item before merge: the config screen needs to guard against overwriting the stored Ceros API key with a masked placeholder on re-save (issue 1). Two smaller non-blocking notes are also included as comments: unvalidated path interpolation in the Function (issue 2), and a discarded `variant` argument in the embed confirmation flow (issue 3).

primeinteger (approved):
Everything looks good to me. Thanks for addressing the feedback from Mitch.

I'll update the App Definition and release these updates to the Ceros app on Contentful Marketplace shortly after.

primeinteger:
Hey @evanmcc92, after merging this PR to update the Production app, I was doing post-release testing and noticed a critical issue:

The app was trying to fetch the app action by the install org, which 404s whenever the app is installed into an org other than the one that manages its App Definition (which is always the case for Marketplace apps).

I made these changes to fix that issue, so now the app works cross-org: https://github.com/contentful/marketplace-partner-apps/pull/8762

PR #8766, opened 2026-09-11
feat(intershop): Add ability to reorder products, drops unimplemented locations + small bug fixes

Hans-Robrecht:
## Purpose

Currently the app stores Intershop products in the order they were selected, so newly selected products always land at the end of the list. Content editors have no way to control the displayed order (e.g. to highlight a promotion) outside of removing the entire product list and defining the entire order again.

This change adds the ability to drag-and-drop the products in the Field location and uses the stored field value product order as the source of truth for display order of the field and the dialogs product basket.

## Approach

Other partner apps such as `ìmagekit`, `cloudinary2` already introduce the ability to drag-and-drop items through dnd-kit and the F36-native card draghandles. We did not want to introduce another drag-and-drop package for this so we followed the same mindset.
Implementation wise we followed https://dndkit.com/react/quickstart/ and https://dndkit.com/react/hooks/use-sortable/.

### Commits

All commits until 12c522a36276087a5d1890c66eed7f4cfcf4d974 are refactor changes where the file was formatted and unused imports were removed. 
The only exception being 11aebacc4d13fda00e1aec59fda4cf66a63d5375 due to installs being blocked by a node mismatch of the current packages. 
We synced our `.nvmrc` file with the latest release from `lts/iron` listed at https://nodejs.org/en/about/previous-releases.

Commits 12c522a36276087a5d1890c66eed7f4cfcf4d974 to 219d8289874f476692af31ab72d64ab505a99252 along with 3de1ffb11fd11c30852be3f97aee7a9fcd1bd10c to 2f3e8b4a7573639cce6b1499b413867f308a5473 contain the sorting logic.

The commits starting from f9d0914461e6b8c9921e789e921def91b9c5ce7e are bug fixes.

### Bug fixes

We focused our effort on small isolated UX changes that shouldn't impact existing application users.

### Product reordering (happy path)

1. User enters an entry
2. Selects various products in the dialog screen (through "Product(s)" or "Change selection" button depending on whether the user already selected products)
3. Sees the products in the product basket according to their selection order
4. Presses "Save" in the dialog
5. Sees the products in the field product list
6. Reorders the products using the drag handles in the user preferred order (e.g. moves the last product to the front)
7. Returns to the dialog to see the product basket display the products in their configured order
8. Refreshes the screen
9. Sees the field product list maintain their preferred order

#### Prevent duplicate product call with same offset f9d0914461e6b8c9921e789e921def91b9c5ce7e

1. User enters an entry
2. Selects various products in the dialog screen (through "Product(s)" or "Change selection" button depending on whether the user already selected products)
3. Opens network tab
4. Scrolls down in the product grid to cause a next product batch fetch
5. Observes a call with `offset=20`
6. Scrolls down in the product grid to cause another product batch fetch
7. Observers a call with `offset=40`

#### Display categories error in field

1. Have an entry with selected categories
2. Change the app configuration with an invalid baseAPI
3. Re-enter the entry
4. Observe an error being shown now

#### Application input in the channel/application selector (field location) updates visually when selecting another application in the dropdown e9fdbd4e1c0aabab25678907c0849cf8eb904c05

1. Requires a configured app with at least 2 applications
2. Enter an entry
3. Select "Change Channel/Application" and confirm the field data wipe
4. Select the "Application" dropdown
5. Select any other application value
6. Observe the input value changes to the selected option value

#### Fresh entry render succeeds when less than 2 channels and applications are configured 2269643af47c7e50ebf169681bb985cd561617d3

1. Requires a configured app with at most 1 channel and at most 1 application
2. Create a fresh entry
3. The user sees the rendered app (previously a `cannot destruct field.getValue()` error was thrown in the console, workaround fix being swapping to another tab and returning since the field value is then initialized).

## Breaking Changes

N/A

This main feature topic was only discussed internally, no public traces are available.

## Deployment

This change removed the boilerplate placeholder locations: `EntryEditor`, `Home`, `Page` and Sidebar`. 
In the case someone is using these locations even though they don't contain actual logic. What'll be the effect?


apps/intershop/src/locations/Field.tsx
```diff
@@ -277,6 +293,10 @@ const Field = () => {
         });
         setChannel(channel);
         setApplication(application);
+        if (channels.length <= 1 && applications.length <= 1) {
```
Hans-Robrecht, line 296: Specifically only matching the scenario where both channels and applications are either not set or only have a single value.
If we were to always set the selected states then the default display of the channel and application selector would be bypassed. 
That logic depends on no selectedChannel or selectedApplication being present when they have multiple options available.

